PR #6, opened 2020-02-10
Feature/leaderboard

jimmyl02:
Added leaderboard component, added utils folder, removed 'uuid-ossp', added goodLeaderboard response


migrations/1581297115194_add-solves.js
```diff
@@ -0,0 +1,11 @@
+exports.up = function (pgm) {
+  pgm.createTable('solves', {
+    id: { type: 'uuid', primaryKey: true },
+    challengeid: { type: 'string', notNull: true },
```
chen-robert: shouldn't challengeid + userid be of type `uuid`?

server/api/leaderboard.js
```diff
@@ -0,0 +1,67 @@
...
+      }
+      // Store which challenges each user solved for later
+      if(!(solves[i].userid in userSolves)){
+        userSolves[solves[i].userid] = [solves[i].challengeid]
+      }else{
+        userSolves[solves[i].userid].push(solves[i].challengeid)
+      }
+    }
+
+    for(let i = 0; i < challenges.getAllChallenges().length; i++){
```
chen-robert: `challenges.getAllChallenges()` should be stored in a local variable? It might be expensive to call

PR #8, opened 2020-02-11
Challenge data integration

chen-robert:
Finished integration of challenge data exports with `server/challenges`

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,20 @@ We have designed rctf with the following attributes in mind:
 * simple integration with redpwn's [CTF deployment framework](https://github.com/redpwn/rdeploy)
 * dynamic scoring
 
+## Deployment
+
+### Problem Data
+
+To build, simply copy the exported directory from [rDeploy](https://github.com/redpwn/rdeploy) to `config.rDeployDirectory` (defaults to `.rdeploy`). 
+
+```bash
+~ $ ls
+rdeploy rctf
+~ $ cd rdeploy
+~/rdeploy $ python main.py build
+~/rdeploy $ cp -r export ../rctf/.rdeploy
+```
+
 ## Installation
 
 ### Database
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -15,6 +15,20 @@ We have designed rctf with the following attributes in mind:
...
+### Problem Data
+
+To build, simply copy the exported directory from [rDeploy](https://github.com/redpwn/rdeploy) to `config.rDeployDirectory` (defaults to `.rdeploy`). 
+
+```bash
+~ $ ls
+rdeploy rctf
+~ $ cd rdeploy
+~/rdeploy $ python main.py build
+~/rdeploy $ cp -r export ../rctf/.rdeploy
```
Arinerron: It seems like this is something we could have rctf do too. A potential cool feature would be an API to auto git pull rdeploy dir, re-export and copy to rctf. Then we could add a webhook to the chal dir to auto update rctf whenever we push

server/challenges/index.js
```diff
@@ -1,23 +1,25 @@
...
-        max: 100
-      },
-      files: ['static/security.txt'],
-      id: 'unique/id-here'
-    }]
+    return challenges
   },
   getChallenge: id => {
-
+    return challenges.filter(a => a.id === id)[0]
```
ginkoid: It would be better if this is a constant-time map lookup

server/challenges/index.js
```diff
@@ -1,23 +1,25 @@
+const config = require('../../config')
+const path = require('path')
+
+let challenges
+
+const resetChallenges = () => {
+  const module = path.join('../../', config.rDeployDirectory, 'config.json')
+
+  delete require.cache[require.resolve(module)]
+  challenges = require(module)
```
ginkoid: `require` will cache the module. If we want to add no-downtime challenge reloading, we would have to use `fs.readFile`.

PR #11, opened 2020-02-12
Improved schema validations

chen-robert:
Refactored `routes.schema` into `routes.schema.body` and `routes.schema.params` for more specific validations


jimmyl02:
Fixed errors in the refactor in submitflag.

PR #18, opened 2020-02-15
passwordless login

ginkoid:


config/index.js
```diff
@@ -1,3 +1,8 @@
 module.exports = {
-  rDeployDirectory: '.rdeploy'
+  rDeployDirectory: '.rdeploy',
+  divisions: {
```
chen-robert: Does this specify the different divisions? If so would we be able to specify multiple eligible divisions?

server/auth/token.js
```diff
@@ -2,12 +2,22 @@ const { promisify } = require('util')
 const crypto = require('crypto')
 
 const randomBytes = promisify(crypto.randomBytes)
-const tokenKey = Buffer.from(process.env.APP_TOKEN_KEY, 'base64')
+const tokenKey = Buffer.from(process.env.RCTF_TOKEN_KEY, 'base64')
```
chen-robert: I think we should use `config.tokenKey` instead of hardcoding environmental variables

PR #19, opened 2020-02-15
dockerize app and db deps

ginkoid:


Dockerfile
```diff
@@ -1,16 +1,20 @@
...
 WORKDIR /app
 
-COPY package.json ./
-COPY yarn.lock ./
+ARG rctf_name
+ENV RCTF_NAME=${rctf_name}
 
-ENV NODE_ENV production
+COPY package.json yarn.lock ./
 RUN yarn
```
chen-robert: Could we combine these lines into ```RUN yarn && yarn build && rm -r node_modules```. I think it makes it clearer what the intent is, install dev dependencies and remove node_modules

docker-compose.yml
```diff
@@ -1,11 +1,45 @@
...
+      - rctf
+    env_file:
+      - .env
+    volumes:
+      - /var/data/rctf-rdeploy:/app/.rdeploy
+    depends_on:
+      - redis
+      - postgres
+    cpus: '1'
+    mem_limit: 500M
```
chen-robert: Do we want to set hard memory limits? Maybe make it configurable in .env?

PR #20, opened 2020-02-15
Added auth tests

chen-robert:
Also added `config.verifyEmail` to configure if emails should be sent. If not, the response will be identical to as if you posted the token to `/verify`

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -1,5 +1,6 @@
 module.exports = {
   rDeployDirectory: '.rdeploy',
+  verifyEmail: true,
   tokenKey: process.env.RCTF_TOKEN_KEY,
   ctfName: process.env.RCTF_NAME,
   origin: process.env.RCTF_ORIGIN,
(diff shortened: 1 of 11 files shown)


server/api/auth-test.js
```diff
@@ -0,0 +1,10 @@
+const { responses } = require('../responses')
+
+module.exports = {
+  method: 'get',
+  path: '/auth/test',
+  requireAuth: true,
+  handler: async () => {
+    return [responses.validToken]
```
ginkoid: If you don't include a `body` in the response, just `return responses.validToken` also works.

config/index.js
```diff
@@ -1,5 +1,6 @@
 module.exports = {
   rDeployDirectory: '.rdeploy',
+  verifyEmail: true,
```
ginkoid: Should we add an env var for this?

chen-robert: I think env variables are for things that aren't configurable, like no sane defaults. For example, a database_url has no sane default. verifyEmail has a sane default though, and we don't want to overbloat `.env`

PR #28, opened 2020-02-16
Login page component

jimmyl02:


client/src/components/login.js
```diff
@@ -0,0 +1,84 @@
...
+          route('/challenges')
+        } else if (resp.kind === 'badTokenVerification' || resp.kind === 'badUnknownUser') {
+          this.setState({
+            errorMessage: resp.message,
+            displayError: true,
+            disabledButton: false
+          })
+        } else {
+          this.setState({
+            errorMessage: 'Unknown response from server, please contact ctf administrator',
```
ginkoid: In this situation, we can just display the `.message` response attribute from the server.

jimmyl02: Good idea but this error message code is temporary so I'll fix it when we decide on how to handle errors.

PR #29, opened 2020-02-16
add leaderboard caching and pagination

ginkoid:


app.js
```diff
@@ -1,6 +1,8 @@
 const path = require('path')
 const express = require('express')
 
+require('./server/leaderboard')
```
chen-robert: Is require with side-effects good practice? Maybe add a function to call like `.start()`

server/cache/leaderboard.js
```diff
@@ -0,0 +1,34 @@
...
+}
+
+const getRange = async ({ start, end }) => {
+  let redisResult = []
+  try {
+    redisResult = await redisLrange('leaderboard', start * 3, end * 3 - 1)
+  } catch (e) {}
+  const result = []
+  for (let i = 0; i < redisResult.length / 3; i++) {
+    result.push([redisResult[i], redisResult[i + 1], parseInt(redisResult[i + 2])])
```
chen-robert: The meaning of `redisValue[i + {1,2}]` should be documented somewhere? Maybe in a comment or a javascript class

server/leaderboard/calculate.js
```diff
@@ -0,0 +1,59 @@
...
+  const solvedChalls = userSolves.get(userId)
+  for (let j = 0; j < solvedChalls.length; j++) {
+    // Add the score for the specific solve loaded fr om the challengeValues array using ids
+    currScore += challengeValues.get(solvedChalls[j])
+  }
+  userScores.push([userId, users[i].name, currScore, userLastSolves.get(userId)])
+}
+
+const sortedUsers = userScores.sort((a, b) => {
+  const scoreCompare = b[2] - a[2]
```
chen-robert: Is there a way to make `b[2]` more informative? It's not really clear what property this is referring to.

jimmyl02: a and b refer to elements in userScores but we could document the type of data stored in the array

ex. 
```
/**
 * @typedef {Object[]} userScores - a list of users and their scores
 * @typedef {string} userScores[0] -  userId
 * @typedef {string} userScores[1] -  username
 * @typedef {string} userScores[2] -  last solve time
 */
```

Issue #30, opened 2020-02-16
Error Handling


jimmyl02:
Currently, error handling for the register and login components is inconsistent. We need to decide on a standard way of handling and displaying the information on the client.

PR #32, opened 2020-02-17
Added generic forms

chen-robert:
Some refactoring of the javascript and created a new generic component. Javascript wrapper functions should return a mapping of errors, the business logic shouldn't be mixed in with the frontend code. 

Resolves https://github.com/redpwn/rctf/issues/30


client/src/api/auth.js
```diff
@@ -14,9 +30,17 @@ export const register = ({ email, name, division }) => {
...
     .then(resp => {
-      if (resp.kind === 'verifyEmail') {
-        localStorage.setItem('token', resp.data.authToken)
-        localStorage.setItem('teamToken', resp.data.teamToken)
+      switch (resp.kind) {
+        case 'verifyEmail':
+          localStorage.setItem('token', resp.data.authToken)
+          localStorage.setItem('teamToken', resp.data.teamToken)
+
+          return route('/profile')
```
jimmyl02: Is this correct? I don't see the verifyEmail response kind; should be "goodVerify"? Also, this is only when the server is not in verifyEmail mode, if response is goodVerifySent, route them to a check your email page?

PR #35, opened 2020-02-17
Added more detailed errors to auth, fixed leaderboard

jimmyl02:
Added check if score is 0 for leaderboard

### DIFF
--- a/server/api/auth-submit.js
+++ b/server/api/auth-submit.js
@@ -41,30 +41,10 @@ module.exports = {
     }
     const userByEmail = await database.auth.getUserByEmail({ email })
     const userByName = await database.auth.getUserByName({ name })
-    if (req.body.register) {
-      // User is trying to register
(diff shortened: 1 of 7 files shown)


server/api/auth-submit.js
```diff
@@ -35,15 +35,32 @@ module.exports = {
...
+    } else {
+      // User is trying to recover
+      if (userByEmail === undefined) {
+        // Could not find account with the provided email
+        return responses.badUnknownEmail
+      }
+      if (userByName === undefined) {
+        // Could not find account with the provided name
+        return responses.badUnknownName
+      }
```
chen-robert: Do we not need the ```javascript
      if (userByEmail.name !== tokenData.name) {
        return responses.badEmailNameMatch
      }
``` check here? It appears in `auth-verify.js`. Also, these checks seem duplicated, we can probably refactor them out into a new utility method somewhere.

PR #36, opened 2020-02-17
Add installation script for Debian-based distros and Arch Linux

Arinerron:
TODO sometime in the future:
* run docker-compose not as root?
* support RHEL derivative distributions

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -149,7 +149,7 @@ read result <&1
 
 if [ "$result" = "y" ]; then
     info "Running 'docker-compose up' in ${INSTALL_PATH}..."
-    docker-compose up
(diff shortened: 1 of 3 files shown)


install/install.sh
```diff
@@ -0,0 +1,157 @@
...
+
+info "Finished installation to ${INSTALL_PATH}."
+
+
+/bin/echo -ne "Would you like to start rCTF now (y/N)? "
+read result <&1
+
+if [ "$result" = "y" ]; then
+    info "Running 'docker-compose up' in ${INSTALL_PATH}..."
+    docker-compose up
```
chen-robert: We probably want `docker-compose up -d` actually so it doesn't exist when the shell terminates.

PR #38, opened 2020-02-17
Improve install script

Arinerron:
* use `--noconfirm` in pacman to avoid requirement for user input
* use `--needed` in pacman to not reinstall packages if up to date
* add `-y` in pacman to refresh package listing before installing dependencies
* add `apt-get update` in apt to refresh package listing before installing dependencies
* silence errors involving systemctl and masked units (docker is masked by default on Debian-like distros)
* have install script fix .env permissions automatically

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -88,6 +88,7 @@ info "Installing dependencies..."
 
 
 if [ "$PACKAGE_MANAGER" = "apt-get" ]; then
+    apt-get update
     apt-get install --yes docker.io docker-compose git
 elif [ "$PACKAGE_MANAGER" = "yum" ]; then
     info "We are about to install docker via https://get.docker.com/. Please follow along the steps to ensure it is configured properly."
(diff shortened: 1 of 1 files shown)


install/install.sh
```diff
@@ -137,6 +138,7 @@ cp .env.example .env
 sed -i.bak "s/RCTF_NAME=.*$/RCTF_NAME=$(echo "$RCTF_NAME"  | sed -e 's/\\/\\\\/g; s/\//\\\//g; s/&/\\\&/g')/g" .env
 sed -i.bak "s/RCTF_TOKEN_KEY=.*$/RCTF_TOKEN_KEY=$(echo "$RCTF_TOKEN_KEY"  | sed -e 's/\\/\\\\/g; s/\//\\\//g; s/&/\\\&/g')/g" .env
 
+chmod 600 .env .env.example
```
chen-robert: Do we want to silently change permissions on `.env`? Might be good to alert the user

Arinerron: ig we could alert the user. I didn't alert because it doesn't seem like a major change that the user needs to be aware of.

chen-robert: Ah ok, I guess its a matter of personal preference. Feel free to merge then

PR #42, opened 2020-02-18
Add rCTF CLI management tool

Arinerron:


install/rctf.py
```diff
@@ -0,0 +1,163 @@
...
+# main
+
+
+if __name__ == '__main__':
+    # parse arguments
+
+
+    parser = argparse.ArgumentParser(description = 'Manage rCTF installations from the CLI')
+    
+    parser.add_argument('--install-path', type = str, default = os.environ.get('RCTF_INSTALL_PATH', os.environ.get('INSTALL_PATH', '/opt/rctf/')), help = 'The path to the rCTF installation to manage')
```
chen-robert: Maybe default should be current working directory? A lot of people will probably git clone and cd into rctf.

install/rctf.py
```diff
@@ -0,0 +1,163 @@
...
+    elif subcommand in ['update', 'upgrade']:
+        if not rctf.upgrade():
+            exit(1)
+
+        logging.info('Successfully updated instance. Run `%s up` to start instance' % sys.argv[0])
+        logging.info('Upgrading %s CLI tool...' % sys.argv[0])
+        
+        # XXX: command injection? i can't think of a practical scenario here
+        # TODO/BUG: either way there is the possibility an admin could have a directory with a name containing a "
+        if not execute('cp install/rctf.py "%s" && chmod + "%s"' % (sys.argv[0], sys.argv[0])):
```
chen-robert: We might as well patch it, shouldn't be too hard to escape shell quotes? see: https://stackoverflow.com/questions/35817/how-to-escape-os-system-calls

install/rctf.py
```diff
@@ -0,0 +1,163 @@
...
+import logging
+
+
+logging.basicConfig(format = '%(message)s', level = logging.DEBUG)
+
+
+# define useful functions
+
+
+def verify_privileges(euid = 0):
```
chen-robert: If the user is in the docker group, they would not need root to execute `docker-compose` commands

install/rctf.py
```diff
@@ -0,0 +1,163 @@
...
+        self.install_path = install_path
+
+
+    def up(self):
+        os.chdir(self.install_path)
+        
+        if not verify_privileges():
+            return False
+
+        if not execute('docker-compose up -d'):
```
chen-robert: We might want to use `docker-compose up -d --build` to rebuild any changes that have been made - similarly for install script

install/rctf.py
Arinerron: Since the rCTF autoinstall script installs to /opt/rctf/, I wanted to prioritize that directory. I think long-term we should default to `.` (current directory) and just `sed` to replace the default value to `/opt/rctf/` from the install script. TODO.

install/rctf.py
Arinerron: Yup, I marked this as a draft because there's a lot more to do before merging. I just didn't want to get caught up in details yet cause there's still a lot to write for this PR.

install/rctf.py
Arinerron: The installation script relies on the user being root atm and since this made to work with it, so does it. We should open an issue to enable both scripts to detect if the current user is in the `docker` group, and require root if not.

install/rctf.py
Arinerron: Thanks, that's a good point.

Arinerron:
Note: I moved the information in `README.md` to the wiki.

install/rctf.py
chen-robert: Made it: https://github.com/redpwn/rctf/issues/43

install/rctf.py
Arinerron: fixed this!

README.md
```diff
@@ -10,76 +10,53 @@ We have designed rCTF with the following attributes in mind:
...
 
-To build, simply copy the exported directory from [rDeploy](https://github.com/redpwn/rdeploy) to `config.rDeployDirectory` (defaults to `.rdeploy`). 
+If you use [rDeploy](https://github.com/redpwn/rdeploy), you can automatically import the problem data using the below command:
 
-```bash
-$ cd rdeploy
-$ python main.py build
-$ cp -r export ../rctf/.rdeploy
 ```
+# rctf deploy /path/to/rdeploy/directory/
```
chen-robert: Maybe this should be `rctf deploy /path/to/problem/directory` and we clone rdeploy for them

PR #44, opened 2020-02-18
Challenge file hosting

chen-robert:
I had to create a custom handler for file downloads in order to fully control the headers. `express.static` didn't work for some reason, overriding my Content-Disposition headers. You should be able to go to a file path such as `http://localhost:3000/static/files/testfile-longhashhere.txt` and get a download of `testfile.txt` (note the hash gets removed).


chen-robert:
Switched to [res.sendFile](http://expressjs.com/en/4x/api.html#res.sendFile) which has better security guarantees. 

> When the root option is provided, the path argument is allowed to be a relative path, including containing ... Express will validate that the relative path provided as path will resolve within the given root option.

PR #50, opened 2020-02-19
Added /challs API

chen-robert:
Added `/challs` and improved tests. Also added config start and end times, and additional checks on `/challs/:id/submit`


chen-robert:
I believe test.serial is needed because we are changing config values. If they were to run in parallel, one config change might affect the other. 

I'll add the starts in data to the response though.

PR #51, opened 2020-02-19
refactor submit into register and recover

ginkoid:


test/integration/auth.js
```diff
@@ -60,24 +60,36 @@ test.serial('when not verifyEmail, succeeds with goodVerify', async t => {
...
+    .send({
+      ...testUser,
+      email: 'non-existent-email' + String(Math.random()) + '@gmail.com'
+    })
+    .expect(responseList.badKnownName.status)
+
+  t.is(resp.body.kind, 'badKnownName')
+})
+
 // TODO: remember too remove test user
```
chen-robert: Unrelated to PR but we can get rid of this `TODO: `?

test/integration/auth.js
```diff
@@ -60,24 +60,36 @@ test.serial('when not verifyEmail, succeeds with goodVerify', async t => {
...
       ...testUser,
-      register: true
+      name: String(Math.random())
     })
     .expect(responseList.badKnownEmail.status)
 
   t.is(resp.body.kind, 'badKnownEmail')
 })
 
+test.serial('duplicate name fails with badKnownEmail', async t => {
```
chen-robert: Do you mean `fails with badKnownName`?

PR #53, opened 2020-02-20
Added /challenges frontend

chen-robert:
Refactored some of the `/challs` api endpoint and finished up frontend code.


client/src/pages/challenges.js
```diff
@@ -0,0 +1,97 @@
...
+    getChallenges()
+      .then(problems => {
+        this.setState({
+          problems
+        })
+      })
+  }
+
+  render ({ classes }, { problems }) {
+    return (
```
ginkoid: Refactoring this into multiple components would be better.

chen-robert: I'll make an issue for this

PR #55, opened 2020-02-20
Added api link for /profile

jimmyl02:
Added getCleanedChallenge() in backend and linked frontend and backend.


client/src/pages/profile.js
```diff
@@ -21,16 +22,53 @@ export default withStyles({
...
+      .then(data => {
+        this.setState({
+          name: data.name,
+          division: data.division,
+          score: data.score.score,
+          teamToken: data.teamToken,
+          solves: data.returnedSolves
+        })
+        let placementStr = String(data.score.place)
+        if (data.score.place >= 11 && data.score.place <= 13) {
```
chen-robert: This feels like it could go into a utility method

server/responses/index.js
```diff
@@ -102,6 +102,14 @@ const responseList = {
   goodChallenges: {
     status: 200,
     message: 'The retrieval of challenges was successful'
+  },
+  goodUserData: {
+    status: 200,
+    message: 'The user was successfully retrieved'
+  },
+  badUserData: {
```
chen-robert: Is it possible to provide a more in-depth reason for why it failed to retrieve? badUserData sounds like the user does not exist

server/api/users-me.js
```diff
@@ -0,0 +1,44 @@
...
+    const user = await db.users.getUserByUserId({ userid: uuid })
+
+    if (user === undefined) return responses.badUserData
+
+    const solves = await db.solves.getSolvesByUserId({ userid: uuid })
+    const teamToken = await auth.token.getToken(auth.token.tokenKinds.team, uuid)
+    const score = await cache.leaderboard.getScore({ id: uuid })
+    const returnedScore = score === null ? 0 : score
+
+    const returnedSolves = []
```
chen-robert: This should just be called `solves`?

client/src/pages/profile.js
jimmyl02: This sounds right, I will move it.

test/integration/profile.js
```diff
@@ -0,0 +1,41 @@
...
+  name: uuidv4(),
+  division: Object.values(config.divisions)[0]
+}
+
+test.serial('succeeds with goodUserData', async t => {
+  config.verifyEmail = false
+
+  let resp = await request(app)
+    .post(process.env.API_ENDPOINT + '/auth/register')
+    .send({
```
chen-robert: Small thing, can't we just do `.send(testUser)` here?

PR #57, opened 2020-02-20
Added leaderboard frontend

chen-robert:
image

### DIFF
--- a/client/src/pages/scoreboard.js
+++ b/client/src/pages/scoreboard.js
@@ -27,7 +27,6 @@ export default withStyles({
   }
 
   render ({ classes }, { scores, division }) {
-    console.log(scores)
     return (
       <div class='row u-center' style='align-items: initial !important'>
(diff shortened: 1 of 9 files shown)


client/src/pages/scoreboard.js
```diff
@@ -0,0 +1,82 @@
...
+    getScoreboard()
+      .then(data => {
+        this.setState({
+          scores: data.leaderboard
+        })
+      })
+  }
+
+  render ({ classes }, { scores, division }) {
+    console.log(scores)
```
ginkoid: This can probably be removed.

client/src/pages/scoreboard.js
```diff
@@ -0,0 +1,82 @@
...
+    return (
+      <div class='row u-center' style='align-items: initial !important'>
```
ginkoid: We should refactor this into multiple components later.

PR #63, opened 2020-02-22
Implemented user profiles for non-private

jimmyl02:
Fixed frontend display for when there are no solves


server/api/users-id.js
```diff
@@ -0,0 +1,52 @@
...
+    const userSolves = await db.solves.getSolvesByUserId({ userid: req.params.id })
+    const score = await cache.leaderboard.getScore({ id: req.params.id })
+    const returnedScore = score === null ? 0 : score
+
+    const solves = []
+
+    userSolves.forEach(solve => {
+      const chall = challenges.getCleanedChallenge(solve.challengeid)
+      // TODO: Should return the challenge point value, currently hard coded
+      solves.push([chall.category, chall.name, 450])
```
ginkoid: It would probably be more clear from an api-design standpoint to use object keys instead of an array.

server/api/users-id.js
```diff
@@ -0,0 +1,52 @@
...
+    })
+
+    return [responses.goodUserData, {
+      name: user.name,
+      division: divisionMap.get(Number(user.division)),
```
ginkoid: Considering everything else in the API uses the numerical division IDs, we should return the numerical ID here and then translate that into the human-readable form on the client.

chen-robert: We'll have to refactor the division code configuration in a later PR, I'll do it in then.

PR #68, opened 2020-02-23
Users api functions

chen-robert:
Implemented PATCH and DELETE on `/users/me`

### DIFF
--- a/server/api/users/update.js
+++ b/server/api/users/update.js
@@ -1,0 +1,33 @@
+const { responses } = require('../../responses')
+const database = require('../../auth')
+const config = require('../../../config')
+
+module.exports = {
+  method: 'patch',
+  path: '/users/me',
+  requireAuth: true,
+  schema: {
+    body: {
+      type: 'object',
+      properties: {
+        name: {
+          type: 'string'
+        },
+        division: {
+          type: 'number',
+          enum: Object.values(config.divisions)
+        }
+      }
+    }
+  },
+  handler: async ({ uuid, name, email, division }) => {
+    await database.auth.updateUser({
+      id: uuid,
+      name,
+      email,
+      division
+    })
+
+    return responses.goodUserUpdate
+  }
+}
--- a/server/api/users/update.js
+++ b/server/api/users/update.js
@@ -1,5 +1,5 @@
 const { responses } = require('../../responses')
-const database = require('../../auth')
+const database = require('../../database')
 const config = require('../../../config')
 
 module.exports = {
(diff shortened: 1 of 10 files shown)


test/util.js
```diff
@@ -0,0 +1,14 @@
+require('ava')
```
jimmyl02: Is this ava requirement needed?

test/integration/auth.js
```diff
@@ -87,6 +85,51 @@ test.serial('duplicate name fails with badKnownName', async t => {
...
+
+  const respUser = resp.body.data.user
+  t.is(respUser.name, nextUser.name)
+  t.is(respUser.email, testUser.email)
+  t.is(respUser.division, nextUser.division)
+
+  testUser.name = respUser.name
+  testUser.email = respUser.email
+  testUser.division = respUser.division
+})
```
jimmyl02: I don't think this code is guaranteed to run if the above tests fail. It might be a good idea to put them above the t.is checks to make sure goodUserDelete test doesn't fail due to this one failing.

test/util.js
chen-robert: `nyc ava` will complain if it's not. I couldn't find a way to exclude the file entirely from nyc

server/api/users/update.js
```diff
@@ -0,0 +1,40 @@
...
+  handler: async ({ uuid, req }) => {
+    const { name, division } = req.body
+
+    const user = await database.auth.updateUser({
+      id: uuid,
+      name,
+      division
+    })
+
+    return [responses.goodUserUpdate, {
```
ginkoid: Is it necessary to return user data here? This seems like a place we could forget to update if we ever add a column to the user table.

server/database/auth.js
```diff
@@ -21,14 +21,18 @@ const ret = {
...
+  },
   makeUser: ({ id, name, email, division, perms }) => {
     return db.query('INSERT INTO users (id, name, email, division, perms) VALUES ($1, $2, $3, $4, $5) RETURNING *',
       [id, name, email, division, perms]
     )
       .then(res => res.rows[0])
   },
   updateUser: ({ id, name, email, division, perms }) => {
-    return ret.getUser({ id })
+    return ret.getUserById({ id })
```
ginkoid: Can we do this in one postgres `UPDATE`?

server/api/users/update.js
chen-robert: The data returning is merely meant as a convenience for testing, I don't think the client would ever rely on it? The client can always get their data from `/users/me`

PR #69, opened 2020-02-23
Improved challenges page

jimmyl02:
Made show solved button functional.
Added global toasts.
Hides challenge on successful submit and displays toast.

### DIFF
--- a/client/src/pages/challenges.js
+++ b/client/src/pages/challenges.js
@@ -3,7 +3,7 @@ import config from '../config'
 import 'linkstate/polyfill'
 import withStyles from '../components/jss'
 
-import { getChallenges, submitFlag } from '../api/challenges'
+import { getChallenges, submitFlag, getPrivateSolves } from '../api/challenges'
 
 export default withStyles({
   frame: {
@@ -18,7 +18,9 @@ export default withStyles({
   state = {
     problems: [],
     values: {},
-    errors: {}
+    errors: {},
+    showSolved: false,
+    solveIDs: []
   }
 
   componentDidMount () {
@@ -28,6 +30,17 @@ export default withStyles({
       .then(problems => {
         this.setState({
           problems
+        })
+      })
+
+    getPrivateSolves()
+      .then(data => {
+        const solveIDs = []
+        data.forEach(solve => {
+          solveIDs.push(solve.id)
+        })
+        this.setState({
+          solveIDs: solveIDs
         })
       })
   }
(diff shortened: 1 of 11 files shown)


client/src/pages/challenges.js
```diff
@@ -30,95 +33,129 @@ export default withStyles({
           problems
         })
       })
+
+    getPrivateSolves()
+      .then(data => {
+        const solveIDs = []
+        data.forEach(solve => {
```
chen-robert: We can use `data.map(solve => solve.id)` here?

client/src/pages/challenges.js
```diff
@@ -30,95 +33,129 @@ export default withStyles({
...
   }
 
-  render ({ classes }, { problems, values, errors }) {
+  handleShowSolvesCheckbox = () => {
+    this.setState({
+      showSolved: !this.state.showSolved
+    })
+  }
+
+  renderProblem = (classes, problem, values, errors) => {
```
chen-robert: Probably out of scope for this PR, but we should refactor this into a new component later

client/src/pages/challenges.js
```diff
@@ -30,95 +33,129 @@ export default withStyles({
...
-        })
+          this.setState({
+            errors: nxt
+          })
+        }
       })
   }
 
-  render ({ classes }, { problems, values, errors }) {
+  handleShowSolvesCheckbox = () => {
```
chen-robert: Is it possible to use `this.linkState` like we do in /login and /register pages?

client/src/pages/challenges.js
```diff
@@ -30,95 +33,129 @@ export default withStyles({
...
-                            }
-
-                          </div>
-                        </div>
-                    }
-                  </div>
-                </div>
-              )
-            })
+            showSolved
```
chen-robert: This should probably not be inline, it's a bit hard to read. Maybe move it above the return statement?

client/src/pages/challenges.js
```diff
@@ -30,95 +33,129 @@ export default withStyles({
...
+              ? problems.map(problem => {
+                return this.renderProblem(classes, problem, values, errors)
+              })
+              : problems.filter(problem => {
+                return solveIDs.indexOf(problem.id) === -1
```
chen-robert: `solveIDs.includes(problem.id)`?

server/api/users/me-solves.js
```diff
@@ -0,0 +1,23 @@
...
+  path: '/users/me/solves',
+  requireAuth: true,
+  handler: async ({ uuid }) => {
+    const userSolves = await db.solves.getSolvesByUserId({ userid: uuid })
+
+    if (userSolves === null) return responses.badUserData
+
+    const returnedSolves = []
+
+    userSolves.forEach(solve => {
```
chen-robert: We can probably use `let ret = userSolves.map(solve => solve.challengeid)` and return ret?

client/src/util/toasts/withToastProvider.js
```diff
@@ -0,0 +1,56 @@
+import React, { useState, useMemo, createPortal } from 'preact/compat'
+
+import ToastContext from './context'
+import Toast from './Toast'
+
+// Create a random ID
+function generateUEID () {
```
ginkoid: We should investigate returning a reference or a `Symbol`, because generating a string ID seems quite flaky.

client/src/pages/challenges.js
jimmyl02: Doesn't seem like linkState works with checkboxes

client/src/pages/challenges.js
jimmyl02: With the current method of rendering each problem, making this its own function is a bit messy. Feel free to change if you think its better.

chen-robert: Pushed a fix moving it above the return statement

client/src/util/toasts/withToastProvider.js
chen-robert: I'll make an issue for this

PR #70, opened 2020-02-25
Refactor configuration setup

chen-robert:
Moved all configuration to `config/`. 

Also added real division loading to `/register`.


.gitignore
```diff
@@ -11,7 +11,7 @@ yarn-error.log*
 /.env
 
 # Configurationg
```
jimmyl02: small typo

PR #72, opened 2020-02-25
Improve install scripts

Arinerron:


chen-robert:
Small note, we should probably replace the `example_postgres_password` with a randomly generated string too. 

https://github.com/redpwn/rctf/blob/master/.env.example#L5-L6

chen-robert:
@Arinerron is this ready for review yet?

Arinerron:
> Small note, we should probably replace the `example_postgres_password` with a randomly generated string too.
> 
> https://github.com/redpwn/rctf/blob/master/.env.example#L5-L6

Will docker-compose "detect" this change?

Arinerron:
@chen-robert I merged master into this branch but am still getting:
```
$ node-pg-migrate --database-url-var=RCTF_DATABASE_URL up
warning Cannot find a suitable global folder. Tried these: "/usr/local, /.yarn"
/bin/sh: 1: node-pg-migrate: not found
error Command failed with exit code 127.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

chen-robert:
@Arinerron As long as we replace it before running `docker-compose up` it should work. We should also add the `--build` flag, like `docker-compose up --build` to rebuild any changes.

Dockerfile
```diff
@@ -13,10 +13,11 @@ COPY client ./client
 COPY config ./config
 COPY public ./public
 COPY preact.config.js .
-RUN yarn && yarn build && rm -r node_modules
```
chen-robert: This was intentional, shouldn't be changed. `yarn build` requires dev dependencies but to reduce image size, we remove them after building. Then we set environment to production and run `yarn` again to reinstall production dependencies for server.

Arinerron: This was just me debugging and trying to figure out what was going on, dw

Arinerron:
@chen-robert Is the config.sh script / Dockerfile functional now?

chen-robert:
Should be, travis is still complaining for some reason. I'll see if I can get someone else to take a look but in the meantime, I'll probably disable it as a requirement for merging

chen-robert:
@Arinerron should I merge this now?

PR #73, opened 2020-02-28
Update profile frontend card

chen-robert:
Added a card for the frontend delete and update profile operations. 

image


client/src/pages/profile.js
```diff
@@ -87,7 +109,30 @@ export default withStyles({
...
+    updateAccount(this.state.updateName, this.state.updateDivision)
+      .then(resp => this.setState({
+        name: resp.user.name,
+        division: divisionMap.get(Number.parseInt(resp.user.division)),
+        disabledButton: false
+      }))
+  }
+
+  handleDelete = e => {
+    const resp = prompt('Please type your team name to confirm: ' + this.state.name)
```
ginkoid: `prompt` blocks all execution the main thread, we should not be using it.

PR #76, opened 2020-03-01
add configurable file logging

ginkoid:


config/server.js
```diff
@@ -10,6 +10,7 @@ const ret = {
   redisUrl: process.env.RCTF_REDIS_URL,
   smtpUrl: process.env.RCTF_SMTP_URL,
   emailFrom: process.env.RCTF_EMAIL_FROM,
+  logFile: process.env.RCTF_LOG_FILE,
```
chen-robert: Should we mount `./data` and modify install script to create a default log file location. And maybe leave it as a comment so its easy for people to toggle?

chen-robert: TODO: Separate PR

PR #80, opened 2020-03-19
Switch to eslint and enforce style

ethanwu10:
Switch from using Standard's CLI to using ESLint with the Standard config for more flexibility in lint rules. Also add lint rules for Preact and ava, and fix (both automatically and manually) as many lint errors as possible.

Outstanding lint errors: client/src/pages/profile.js calls `setState()` inside of `componentDidUpdate`

Fixes:
- rewrote `useMemo` logic in toast-related code to create the bound functions inside `useMemo()`, and use the memoized functions exclusively in the component
- test/unit/util.js was being ignored - fixed
- test/util.js moved to test/_util.js to be implicitly ignored by ava instead of using an explicit ignore rule
- use `t.true` instead of `t.truthy` on boolean values
- move `test.after` hooks to be declared before tests are declared, ensuring they are actually run


client/src/util/toasts/withToastProvider.js
```diff
@@ -7,22 +7,23 @@ import Toast from './Toast'
...
 
   return first + second
 }
 
 function withToastProvider (Component) {
   function WithToastProvider (props) {
     const [toasts, setToasts] = useState([])
-    const add = content => {
-      const id = generateUEID()
+    const providerValue = useMemo(() => ({
```
ethanwu10: Change this `useMemo` call to `useCallback`, since `useCallback(x, deps)` is equivalent to `useMemo(() => x, deps)` (see [React docs](https://reactjs.org/docs/hooks-reference.html#usecallback))

PR #83, opened 2020-03-20
Create pluggable provider system for challenges

### LINKED ISSUE #54: Refactor `/challs`
There is a lot of code in the `/challs` component. We should refactor this into multiple smaller components. 

https://github.com/redpwn/rctf/blob/master/client/src/pages/challenges.js#L34

ethanwu10:
Refactor loading of challenges to use a flexible provider system in preparation for implementing alternative sources for challenge data.

The current API surface is a constructor for the provider object, which is provided a callback function for when the challenges change (it is called with the full new challenges array), and an options object lifted verbatim from the config. The created object has methods to force an update and to clean-up (which is currently unused)

Note:
- eslint thinks that having a constructor with side-effects (in this case, setting up a background update that will trigger the callback) is a bad idea - thoughts on this API?


server/providers/challenges/rdeploy-blob/index.js
```diff
@@ -0,0 +1,52 @@
...
+      ? config.rDeployDirectory
+      : options.rDeployDirectory
+    this._interval = setInterval(() => this._update(), this._updateInterval)
+    this._update()
+  }
+
+  _update () {
+    const configPath = path.join(__dirname, '../../../../', this._rDeployDirectory, 'config.json')
+
+    fs.readFile(configPath, (err, data) => {
```
ginkoid: We could probably make this more succinct with async/await.

server/providers/challenges/rdeploy-blob/index.js
```diff
@@ -0,0 +1,52 @@
...
+class RDeployBlobProvider {
+  constructor ({ options, onUpdate }) {
+    options = {
+      updateInterval: 60 * 1000,
+      useGlobalRDeployDirectory: true,
+      // rDeployDirectory: ''
+      ...options
+    }
+
+    this._onUpdate = onUpdate
```
ginkoid: Instead of only supporting one update listener, the plugin could emit events on an EventEmitter.

server/providers/challenges/rdeploy-blob/index.js
ethanwu10: We should pull fs-extra for promisified fs calls then

ethanwu10:
We should have an interface definition for the plugins, but I'm not sure where to define them. Adding them in `server/challenges/` would probably end up with a circular dependency...

chen-robert:
If we did it in a separate file, e.g. `server/challenges/Provider.ts`, and the only export is an interface, there wouldn't be any circular dependencies because the interface wouldn't have any dependencies.

server/challenges/Challenge.ts
```diff
@@ -0,0 +1,30 @@
+export interface Points {
```
chen-robert: We should rename this file to types.ts which [is the standard](https://stackoverflow.com/questions/36633033/how-to-organize-typescript-interfaces).

PR #87, opened 2020-03-20
Installation db mibrate wait for postgres startup

chen-robert:
This script, `install/docker-migrate.sh` should wait for the postgresql instance to startup before attempting to migrate.


Arinerron:
build failed because of broken install script (fixed in other commit), merging...

PR #89, opened 2020-03-20
Add shellcheck CI job

ethanwu10:
Run shellcheck in CI to lint all shell scripts for issues and POSIX compliance

Currently, the shellcheck job is failing, however these issues should be fixed by #88


Arinerron:
CI checks failed because my script is buggy. Working as intended. Thanks.

PR #95, opened 2020-03-21
fix(api): remove race in flag submission

ginkoid:


server/api/submitflag.js
```diff
@@ -44,20 +45,29 @@ module.exports = {
...
+
+    if (!crypto.timingSafeEqual(bufSubmittedFlag, bufCorrectFlag)) {
+      return responses.badFlag
+    }
+
+    try {
+      await db.solves.newSolve({ id: uuidv4(), challengeid: challengeid, userid: uuid, createdat: new Date() })
+      return responses.goodFlag
+    } catch (e) {
+      if (e.constraint === 'uq') {
```
ethanwu10: might want to add a comment, but the rest of the code makes it fairly self explanatory

PR #96, opened 2020-03-21
Add typescript support

chen-robert:
Moved all of the build directories into `dist/`. Also added loose typescript integration which allows for the use of Javascript files. This means that we can incrementally migrate instead of doing it all at once.


.eslintignore
```diff
@@ -1,2 +1,3 @@
-# Ignore build
+# Ignore build folders
 /build
+/dist
```
ethanwu10: missing newline at eof

.eslintrc.js
```diff
@@ -1,4 +1,5 @@
 module.exports = {
+  parser:  '@typescript-eslint/parser',
```
ethanwu10: move this to server/ .eslintrc

.eslintrc.js
```diff
@@ -17,6 +18,7 @@ module.exports = {
     sourceType: 'module'
   },
   plugins: [
+    "typescript"
```
ethanwu10: ditto

client/build.sh
```diff
@@ -4,3 +4,6 @@ rm -r build/ssr-build build/manifest.json build/polyfills* build/push-manifest.j
 mv build/*.js* build/assets
 mv build/assets build/static
 cp -r public/* build
+
+rm -rf dist/build
+mv build dist/
```
ethanwu10: missing newline at eof

package.json
```diff
@@ -5,16 +5,19 @@
...
-    "dev:client": "preact watch --src client/src --template client/index.html",
-    "dev": "nodemon index.js",
-    "test": "nyc --skip-full ava",
+    "build:client": "sh client/build.sh",
+    "build:ts": "tsc && yarn copy-static",
+    "build": "yarn build:ts && yarn build:client",
+    "watch:client": "preact watch --src client/src --template client/index.html",
+    "watch:server": "nodemon dist/index.js",
+    "watch:ts": "tsc -w",
+    "dev": "yarn build:ts && concurrently -k -t \"HH:mm:ss.SSS\" -p \"[{time}]\" -c \"cyan,green,red\" \"yarn watch:client\" \"yarn watch:server\" \"yarn watch:ts\"",
```
ethanwu10: maybe 1) change to a `watch:*` and 2) have output labeled by which watch job in addition to time? (2 might not be necessary since they're already different colors)

package.json
```diff
@@ -26,29 +29,34 @@
     "ajv": "^6.11.0",
     "commitizen": "^4.0.3",
     "email-validator": "^2.0.4",
-    "eslint-plugin-ava": "^10.2.0",
     "express": "^4.17.1",
     "mustache": "^4.0.0",
     "node-pg-migrate": "^4.2.2",
     "nodemailer": "^6.4.6",
+    "nyc": "^15.0.0",
```
ethanwu10: why is this a full dependency now?

tsconfig.json
```diff
@@ -0,0 +1,18 @@
...
+    "inlineSourceMap": true,
+    "inlineSources": true,
+    "allowJs": true,
+    "outDir": "dist",
+    "baseUrl": "."
+  },
+  "include": [
+    "src/**/*", "index.js", "app.js"
+  ]
+}
```
ethanwu10: missing newline at eof

package.json
chen-robert: `dev` is more declarative cause it's the thing you're gonna be using for development.

.eslintrc.js
chen-robert: Don't we need this for `index.js` too?

package.json
ethanwu10: no I meant replace the watch: in the concurrently call

.eslintrc.js
ethanwu10: you're right... maybe move app and index to inside server/?

PR #97, opened 2020-03-21
add support for redis-based rate limits on the server

ginkoid:


package.json
```diff
@@ -18,8 +18,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "yarn lint",
-      "prepare-commit-msg": "exec < /dev/tty && git cz --hook || true"
```
ethanwu10: split this out of this pr

test/unit/rateLimit.js
```diff
@@ -0,0 +1,34 @@
...
+    limit: 1,
+    userid
+  })
+  const result = await timeouts.checkRateLimit({
+    type: timeouts.types.UPDATE_PROFILE,
+    duration: 1000 * 10,
+    limit: 1,
+    userid
+  })
+  t.is(result.ok, false)
```
ethanwu10: use `t.false`?

client/src/api/profile.js
```diff
@@ -25,5 +26,12 @@ export const updateAccount = (name, division) => {
...
     division: Number.parseInt(division)
   })
-    .then(resp => resp.data)
+    .then(resp => {
+      switch (resp.kind) {
+        case 'badRateLimit':
+          toasts.useToast().add(resp.message)
+          return null
+      }
+      return resp.data
```
ethanwu10: have the success case (returning data that should only be valid when the response type is what you expect) in the switch and validate type; throw error in default case?

server/cache/timeouts.js
```diff
@@ -0,0 +1,48 @@
...
+  end
+`)
+
+const types = {
+  FLAG: 'FLAG',
+  UPDATE_PROFILE: 'UPDATE_PROFILE'
+}
+
+module.exports = {
+  /*
```
chen-robert: we should update this comment to reflect new behavior of returning an object

PR #102, opened 2020-03-21
Allow querystring parameters as an object in request

ethanwu10:
Refactor `request()` to automatically put the contents of `data` into the querystring of a request if the request method is `GET`. This implementation currently will add extraneous `?`s to the request URL
when no querystring parameters are provided, however it appears that it is stripped off by the browser.

### DIFF
--- a/client/src/api/util.js
+++ b/client/src/api/util.js
@@ -11,14 +11,15 @@ export const request = (method, endpoint, data) => {
   let qs = ''
   if (method === 'GET' && data) {
     // encode data into the querystring
-    qs = Object.keys(data)
+    // eslint-disable-next-line prefer-template
+    qs = '?' + Object.keys(data)
       .filter(k => data[k] !== undefined)
       .map(k => `${k}=${encodeURIComponent(data[k])}`)
       .join('&')
   } else {
     body = data
   }
-  return fetch(`${config.apiEndpoint}${endpoint}?${qs}`, {
+  return fetch(config.apiEndpoint + endpoint + qs, {
     method,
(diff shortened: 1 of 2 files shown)


client/src/api/util.js
```diff
@@ -7,13 +7,24 @@ export const relog = () => {
 }
 
 export const request = (method, endpoint, data) => {
-  return fetch(config.apiEndpoint + endpoint, {
+  let body = null
+  let qs = ''
+  if (method === 'GET' && data) {
```
chen-robert: Why not do `qs='?'` here and replace "fetch(`${config.apiEndpoint}${endpoint}?${qs}`" with "fetch(`${config.apiEndpoint}${endpoint}${qs}`". This way we don't have the extra "?"

client/src/api/util.js
```diff
@@ -7,13 +7,24 @@ export const relog = () => {
...
+    body = data
+  }
+  return fetch(`${config.apiEndpoint}${endpoint}?${qs}`, {
     method,
     headers: {
       'Content-Type': 'application/json',
       Authorization: `Bearer ${localStorage.getItem('token')}`
     },
-    body: JSON.stringify(data)
+    body: body && JSON.stringify(body)
```
chen-robert: We might want to do strict type checking against `null` incase some false-y values get passed in as data to a POST request?

ethanwu10: `body` should be an object though, since we're `JSON.stringify`ing it, and objects are truthy, so this should be fine provided the function is called correctly

PR #104, opened 2020-03-22
Add scoreboard pagination

ethanwu10:
image
image
image

(actual build uses 25, 50, and 100 as page size options, defaulting to 100)


ethanwu10:
image
image
(actual build does not have 1 and 2 as page size options)

ethanwu10:
Currently there is no UI indicators for when scoreboard data is loading, and the ranking number columns will update on a page change even when the data for that page has not loaded (and will display wrong ranking numbers for the same teams on the previous loaded page until data has loaded). Do we want to fix this now, or just merge this first?

chen-robert:
@ethanwu10 I think make an issue and we can fix it later. This PR is quite big already

ethanwu10:
We may also want the "go to my team" button to scroll the user's team into view, but I think that's out of scope of this PR (and probably better implemented together with #109)

PR #108, opened 2020-03-22
Return division numbers instead of text in user profile API

ethanwu10:
**BREAKING CHANGE**: division in the returned user object is now the numeric ID of the division instead of its text name


ethanwu10:
Closed by #104

PR #112, opened 2020-03-24
feat(api): add server support for leaderboard graphs

ginkoid:


server/leaderboard/calculate.js
```diff
@@ -1,70 +1,133 @@
...
+  })
+
+  parentPort.postMessage({
+    leaderboards: output
+  })
+} else {
+  const nextSample = getNextSample()
+  let sample = Date.now()
+  let isSample = false
+  if ((nextSample - Date.now()) < config.leaderboardUpdateInterval) {
```
ethanwu10: We should decide whether or not to calculate a new sample based on the time of the previous sample instead of trying to guess based on the current time and when the next sample "should" be.

server/leaderboard/calculate.js
```diff
@@ -1,70 +1,139 @@
...
-challengeValues.forEach((score, id) => [
-  challengeScores.push(id, score)
-])
+  const { challengeValues, userScores } = calculateScores(sample)
 
-parentPort.postMessage({
-  leaderboard: sortedUsers,
-  challengeScores
-})
+  const sortedUsers = userScores.sort(userCompare).map((user) => user.slice(0, 4))
```
chen-robert: We might want a more formal definition for the `user` object, is it possible to use Typescript for this?

server/leaderboard/calculate.js
```diff
@@ -1,70 +1,139 @@
 const { workerData, parentPort } = require('worker_threads')
 const utilScores = require('../util/scores')
+const { calcSamples, getPreviousSample } = require('./samples')
 
-const { solves, users, allChallenges } = workerData
+const {
```
chen-robert: This is a lot of destructuring, it might be cleaner to refer to the variables as `workerData.XXX`.

server/leaderboard/calculate.js
ginkoid: We can do this in another PR.

server/leaderboard/calculate.js
ginkoid: The variables destructured from `workerData` are used a lot. It's probably less messy to just destructure in in the beginning.

PR #125, opened 2020-03-25
Reimplement toasts

ethanwu10:
Reimplement toasts with a more sensible structure, using a provider component for context instead of wrapping a component, and cleaning up implementation.


client/src/components/toast.js
```diff
@@ -0,0 +1,86 @@
...
+    zIndex: 9999,
+    width: '320px'
+  }
+}, ({ classes, ...props }) => <div class={classes.container} {...props} />)
+
+export function ToastProvider ({ children }) {
+  const [toasts, setToasts] = useState([])
+
+  const remove = useCallback((id) => {
+    setToasts(toasts => toasts.filter(t => id !== t.id))
```
ginkoid: Would it be possible to use a `Map` for the toasts so that we dont have to do this `O(n)` removal?

client/src/pages/profile.js
```diff
@@ -342,4 +326,28 @@ export default withStyles({
...
+    marginRight: '25px'
+  },
+  form: {
+    '& button': {
+      margin: 0,
+      lineHeight: '20px',
+      padding: '10px',
+      float: 'right'
+    },
+    padding: '0 !important'
```
ginkoid: Could we add an option to the `Form` component instead of overriding the padding using `!important`?

chen-robert: Created an issue, seems out of scope of this PR

client/src/components/toast.js
chen-robert: Probably not a super big issue, converting to starter issue for someone to tackle and get more familiar with codebase

PR #133, opened 2020-03-28
Update preact-cli, clean up client build

### LINKED ISSUE #131: Rewire client build and use preact-cli@3.0.0-rc.*
The client build is currently dependent on a shell script, and is running on an older version of `preact-cli` (the starter projects created by `preact-cli` all use `preact-cli@3.0.0-rc.6`, so we should probably also be using the rc versions). Given the refactoring that needs to be done to the webpack configuration to do this upgrade, we should also fix some other issues / annoyances with our client-side build system.

**Goals**
- no shell script to build - build everything through `preact build` (with flags set)
- use SVG sprite sheets for our SVG loader to avoid duplicating inline SVG - most likely we will use [this loader](https://github.com/JetBrains/svg-sprite-loader), and [an example](https://github.com/JetBrains/svg-sprite-loader/blob/master/examples/custom-runtime-generator/svg-to-icon-component-runtime-generator.js) is included for generating React components automatically (which is what happens with our current config using `preact-svg-loader`)
- use preact-cli's builtin API proxying capabilities in dev mode to avoid needing to have the client access the API cross-origin in dev mode
- fix serving of static files on the server to not require physically moving build outputs on disk (technically a sub-point of removing the shell script)

Upgrading preact-cli (and thus webpack) also will solve our issues with portals [crashing on unmount](https://github.com/redpwn/rctf/blob/438b6adf32dbb491efccc961f3a41d1c8ff8a414/client/src/components/modal.js#L34), because this was caused by Webpack resolving the import to the main `'preact'` module differently from preact's submodules (`'preact/compat'`), which are incompatible due to Preact's output setup and mangler in their build.

ethanwu10:
Update preact-cli to 3.0.0-rc.10 (since the RC versions are being used by the default template) and update the HTML template to match.

**Build changes**
`client/build.sh` has been removed, with the build now being done entirely by preact-cli. The static build output is no longer being served under `/static`, and is now served directly at the root. `preact watch` now also proxies API requests; it is no longer necessary to change `config/client.js` to have API endpoints point to `localhost:3000`.

**Source structure changes**
Move `client/src/pages` to `client/src/routes` to enable route splitting, move icons to within the `client/src` tree, and move contents of `public` directory to the appropriate places within the `client/src` tree.

**Misc**
PurgeCSS has been updated to the latest version.

Prerendering is still disabled due to build failures from sourcemap resolution (we might want to look into this in a separate issue).

This PR also contains small bugfixes in components and routes that were discovered in the process of migration:
- `/profile` now will load data on initial mount instead of only on a state change
- The scoreboard no longer passes `false` as a `ref` (which crashes the app)
- Modals now use portals

Will close #131 
Still TODO:
- [ ] SVG sprite sheets


client/src/routes/profile.js
```diff
@@ -161,6 +146,26 @@ class Profile extends Component {
     }
   }
 
+  static getDerivedStateFromProps (props, state) {
+    if (props.uuid !== state.uuid) {
+      return {
+        uuid: props.uuid,
```
ginkoid: It looks like we're just syncing `state.uuid` for every change in `props.uuid`, which is [an antipattern](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html).

chen-robert: Isn't our code very similar [to the documentation](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#alternative-1-reset-uncontrolled-component-with-an-id-prop)? The only difference is the variable naming. 

In the future, we might want to refactor this with `key={props.uuid}` as recommend though?

ginkoid: Oh true, my bad.

chen-robert:
@ethanwu10 Are we ready to merge this?

ethanwu10:
We can merge this now and make the sprite sheet a separate issue / PR.

PR #144, opened 2020-03-29
client graph support

ginkoid:


client/index.html
```diff
@@ -11,7 +11,7 @@
 		<link href="https://fonts.googleapis.com/css?family=Montserrat:400,700" rel="stylesheet">
 		<% preact.headEnd %>
 	</head>
-	<body>
+	<body style="overflow-x: hidden;">
```
ethanwu10: We should probably apply this as a global style sheet

client/src/components/graph.jsx
```diff
@@ -0,0 +1,207 @@
...
+import { getGraph } from '../api/scoreboard'
+import withStyles from './jss'
+
+const width = window.innerWidth
+const height = 400
+const stroke = 5
+const axis = 20
+const axisGap = 10
+const day = 24 * 60 * 60 * 1000
+const viewBox = `${-stroke - axis} ${-stroke} ${width + stroke * 2 + axis} ${height + stroke * 2 + axis + axisGap}`
```
ethanwu10: `width` and thus `viewBox` should be calculated within the component so we can handle window resizes (recalculate & redraw on the resize event)

client/src/components/graph.jsx
```diff
@@ -0,0 +1,207 @@
...
+      tooltipContent: ''
+    })
+  }
+
+  render({ classes }, { polylines, labels, tooltipContent, tooltipX, tooltipY, tooltipFlipX }) {
+    return (
+      <div class={`frame ${classes.root}`}>
+        <div class='frame__body'>
+          <svg viewBox={viewBox}>
+            <Fragment>
```
ethanwu10: These fragments are not doing anything

client/src/components/graph.jsx
```diff
@@ -0,0 +1,207 @@
...
+              {polylines.map(({ points, color, name, currentScore }, i) => (
+                <polyline
+                  key={i}
+                  stroke-linecap='round'
+                  stroke={color}
+                  stroke-width={stroke}
+                  fill='transparent'
+                  points={points}
+                  pointer-events='stroke'
+                  onMouseOver={this.handleTooltipIn(`${name} - ${currentScore} points`)}
```
ethanwu10: For performance, it would be best not to create a new handler function every render

client/src/routes/scoreboard.js
```diff
@@ -111,6 +112,9 @@ const Scoreboard = withStyles({
 
   return (
     <div class='row u-center' style='align-items: initial !important'>
+      <div class='col-12'>
```
ethanwu10: Should this be in its own row?

client/src/components/graph.jsx
ethanwu10: The fragments prevent key conflicts, so keeping them is fine

client/src/components/graph-line.jsx
```diff
@@ -0,0 +1,11 @@
+import { memo } from 'preact/compat'
+
+export default memo(({ onTooltipIn, name, currentScore,...rest }) => (
```
ethanwu10: missing space between `,` and `...`

client/src/components/graph.jsx
```diff
@@ -65,6 +63,9 @@ export default withStyles({
     padding: '5px 10px',
     borderRadius: '5px',
     margin: '5px'
+  },
+  '@global body': {
+    overflowX: 'hidden'
```
ethanwu10: imo this should be in `index.js` or some other top-level location

client/src/components/graph-line.jsx
```diff
@@ -0,0 +1,11 @@
+import { memo } from 'preact/compat'
```
ethanwu10: Given that this component is very tightly coupled to Graph (and takes input generated by a function from that component), it would make more sense to define this in `graph.js`, especially since it is not reusable

client/src/components/graph.jsx
```diff
@@ -1,16 +1,15 @@
 import { Component, Fragment } from 'preact'
```
ethanwu10: This file should have a `.js` extension

client/src/components/graph.js
```diff
@@ -0,0 +1,229 @@
+import { Component, Fragment } from 'preact'
+import { getGraph } from '../api/scoreboard'
+import withStyles from './jss'
+import { memo } from 'preact/compat'
+
+const height = 400
+const stroke = 5
+const axis = 20
+const axisGap = 10
```
ethanwu10: We may want to make these props, `height` could be useful to have controllable.

ethanwu10: An alternative is to have the height be calculated automatically based on the size of the containing element (plotly has an option to do this), but that could become complex to implement.

client/src/components/graph.js
```diff
@@ -0,0 +1,229 @@
...
+              stroke-width={stroke}
+            />
+          </svg>
+        </div>
+        {tooltipContent && (
+          <div
+            class={classes.tooltip}
+            style={{
+              left: tooltipX,
+              top: tooltipY
```
ethanwu10: `transform` should be animated instead to avoid re-layout and re-paint

PR #169, opened 2020-04-03
feat(api): add initial server-side ctftime integration

ginkoid:


Arinerron:
@kyprizel we love you!

Arinerron:
btw @kyprizel ginkoid didn't leave a description but here's the explanation from Discord:
plzfix

you should change this

kyprizel:
Actually CTFtime do use Github, Gmail, Twitter and Facebook as auth providers.
So all of them do check emails and we can trust them. But yes - then user can change email in profile without limitations. Nice catch - i'll try to fix it.

Thanks for this integration!

PR #172, opened 2020-04-05
allow update, remove of auth methods to existing users

ginkoid:


config/server.js
```diff
@@ -8,7 +8,7 @@ const ret = {
   },
   rDeployDirectory: '.rdeploy',
   rDeployFiles: 'files',
-  verifyEmail: false,
+  verifyEmail: true,
```
chen-robert, line 11: Should this be the default?

server/api/users/me-auth/email/patch.js
```diff
@@ -0,0 +1,58 @@
...
+    const verifyUuid = uuidv4()
+    await cache.login.makeLogin({ id: verifyUuid })
+    const verifyToken = await auth.token.getToken(auth.token.tokenKinds.verify, {
+      verifyId: verifyUuid,
+      kind: 'update',
+      userId: user.id,
+      email
+    })
+
+    await util.email.sendVerification({
```
chen-robert, line 50: Possible race condition where we send two verification emails from TOCTOU? No real impact though and not sure if it's possible to avoid.

server/api/auth/recover.js
```diff
@@ -41,7 +41,7 @@ module.exports = {
     await cache.login.makeLogin({ id: verifyUuid })
     const verifyToken = await auth.token.getToken(auth.token.tokenKinds.verify, {
       verifyId: verifyUuid,
-      register: false,
+      kind: 'recover',
```
chen-robert, line 44: Should we encode the kind as a typescript enum?

server/api/auth/verify.js
```diff
@@ -27,13 +27,13 @@ module.exports = {
       return responses.badTokenVerification
     }
 
-    if (tokenData.register) {
+    if (tokenData.kind === 'register') {
```
chen-robert, line 30: re: typescript enum

server/api/users/me-auth/ctftime/patch.js
```diff
@@ -0,0 +1,34 @@
+const { responses } = require('../../../../responses')
+const database = require('../../../../database')
+const auth = require('../../../../auth')
+
+module.exports = {
+  method: 'patch',
```
ethanwu10, line 6: This should be an HTTP `PUT`, since it can also create a new ctftime auth, and it is already idempotent.

server/api/users/me-auth/ctftime/patch.js
```diff
@@ -0,0 +1,34 @@
...
+          type: 'string'
+        }
+      },
+      required: ['ctftimeToken']
+    }
+  },
+  handler: async ({ req, user }) => {
+    const ctftimeData = await auth.token.getData(auth.token.tokenKinds.ctftimeAuth, req.body.ctftimeToken)
+    if (ctftimeData === null) {
+      return responses.badToken
```
ethanwu10, line 23: We should differentiate between a bad auth token and some other token that is sent being bad, since currently on the client we interpret a `badToken` as meaning that the auth token is bad then force the user to login again.

server/api/users/me-auth/email/delete.js
```diff
@@ -0,0 +1,27 @@
...
+    try {
+      result = await database.auth.removeEmail({ id: user.id })
+    } catch (e) {
+      if (e.constraint === 'require_email_or_ctftime_id') {
+        return responses.badZeroAuth
+      }
+      throw e
+    }
+    if (result === undefined) {
+      return responses.badEmailNoExists
```
ethanwu10, line 23: `badEmailDoesNotExist`?

server/api/users/me-auth/email/patch.js
```diff
@@ -0,0 +1,58 @@
...
+const emailValidator = require('email-validator')
+const config = require('../../../../../config/server')
+const { responses } = require('../../../../responses')
+const cache = require('../../../../cache')
+const util = require('../../../../util')
+const auth = require('../../../../auth')
+const database = require('../../../../database')
+
+module.exports = {
+  method: 'patch',
```
ethanwu10, line 11: This should also be a `PUT`

server/api/auth/recover.js
ethanwu10: Ideally the parameter to getToken should be a single discriminated union so we also have the contents of the token defined by an interface, however I’m not sure how compatible that is with pure JS code so that might need to wait for when we fully convert auth to TS

PR #190, opened 2020-04-13
Add admin challenge data API

chen-robert:
Added admin challenge API. Still missing authentication control + linking to expose endpoints


server/providers/challenges/rdeploy-blob/index.ts
```diff
@@ -56,6 +58,22 @@ class RDeployBlobProvider extends EventEmitter implements Provider {
     this._update()
   }
 
+  updateChallenge (chall: Challenge): void {
+    let updated = false
```
ethanwu10, line 62: Maybe rename this variable to `challengeExists` or similar?

server/challenges/Provider.ts
```diff
@@ -1,6 +1,14 @@
 import { EventEmitter } from 'events'
+import { Challenge } from './types'
 
 export interface Provider extends EventEmitter {
   forceUpdate (): void;
+  /*
```
ethanwu10, line 6: (style nitpick) we should add a blank line between functions

chen-robert: I feel like style concerns should be formalized in `yarn lint` or ignored.

PR #194, opened 2020-04-13
Cleanup frontend

chen-robert:
Restyled the graph and other small changes to make frontend cleaner.

### DIFF
--- a/client/src/components/graph.js
+++ b/client/src/components/graph.js
@@ -5,20 +5,21 @@ import withStyles from './jss'
 import { memo } from 'preact/compat'
 
 const height = 400
-const stroke = 5
+const stroke = 2
 const axis = 20
-const axisGap = 10
+const axisGap = 20
 const day = 24 * 60 * 60 * 1000
 
 const timeToX = ({ minX, maxX, time, width }) => {
   return (time - minX) / (maxX - minX) * width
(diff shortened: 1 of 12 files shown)


client/src/components/graph.js
```diff
@@ -5,20 +5,21 @@ import withStyles from './jss'
...
-const colors = ['#f44336', '#e91e63', '#9c27b0', '#3f51b5', '#2196f3', '#00bcd4', '#009688', '#4caf50', '#8bc34a', '#ffeb3b', '#ff9800', '#ff5722']
-
 const uuidToColor = (uuid) => {
-  const uuidInt = parseInt(uuid.slice(-12), 16)
-  return colors[uuidInt % colors.length]
+  const uuidInt = parseInt(uuid.slice(-12), 16) + 1
+  // Random primes
+  const g = uuidInt * 89 % 128
+  const b = uuidInt * 53 % 128
+  return `rgba(186, ${g}, ${b}, 0.8)`
```
ethanwu10, line 22: This could end up producing some nasty colors; should we use hsla and just vary hue based on UUID?

chen-robert: We'll probably have to do some more experimenting, I'll create an issue for this.

PR #196, opened 2020-04-13
Add markdown parsing of challenge descriptions

chen-robert:


ethanwu10:
We may wish to use [preact-markup](https://bundlephobia.com/result?p=preact-markup@2.0.0) to render the html output by snarkdown; it will sanitize the html.

client/src/routes/challenges.js
```diff
@@ -98,7 +106,7 @@ export default withStyles({
 
     const error = errors[problem.id]
     const hasError = error !== undefined
-
+    console.log(snarkdown(problem.description))
```
ethanwu10, line 109: This shouldn't be here

package.json
```diff
@@ -36,6 +36,7 @@
     "nodemailer": "^6.4.6",
     "pg": "^8.0.2",
     "redis": "^3.0.2",
+    "snarkdown": "^1.2.2",
```
ethanwu10, line 39: This should be a devdep, because we are currently only using it on the client

package.json
```diff
@@ -74,6 +75,7 @@
     "pascal-case": "^3.1.1",
     "preact": "^10.4.0",
     "preact-cli": "^3.0.0-rc.10",
+    "preact-markup": "2.0.0",
```
ethanwu10, line 78: We probably want to specify this as `^2.0.0`

PR #200, opened 2020-04-15
Add ctftime login and registration frontend

ginkoid:


client/src/routes/ctftime-callback.js
```diff
@@ -0,0 +1,16 @@
...
+    window.opener.postMessage({
+      kind: 'ctftimeCallback',
+      state: this.props.state,
+      ctftimeCode: this.props.code
+    })
+    window.close()
+  }
+
+  render () {
+    return <span />
```
ethanwu10, line 14: This could also return `null`, but this isn't very important

client/src/routes/login.js
```diff
@@ -42,17 +48,37 @@ export default withStyles({
     }
   }
 
-  render (props, { teamToken, errors, disabledButton }) {
-    const { classes } = props
+  render ({ classes }, { teamToken, errors, disabledButton, ctftimeToken }) {
+    if (ctftimeToken) {
+      return <CtftimeAdditional ctftimeToken={ctftimeToken} />
```
ethanwu10, line 53: Ideally this would redirect to `/registration` while passing the token through using history state, however preact-router does not support attaching state (in JS) to a history entry. We can still implement this by implementing a state store using context and invalidate on any route change, which is enough to implement this.

client/src/components/ctftime-button.js
```diff
@@ -0,0 +1,66 @@
...
+    this.oauthState = openPopup()
+  }
+
+  render ({ classes }, { error }) {
+    return (
+      <div>
+        <button class={classes.ctftimeButton} onClick={this.handleClick}>
+          <Ctftime />
+        </button>
+        {error && (
```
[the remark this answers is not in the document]
ethanwu10: We should render these In a toast.

client/src/routes/login.js
chen-robert: Documented in https://github.com/redpwn/rCTF/issues/201, out of scope for this PR though

PR #203, opened 2020-04-19
Admin frontend

chen-robert:
File uploads still not working; have to rework that still. 

The default provider does not persist changes, only stores in memory.


client/src/components/admin/problem.js
```diff
@@ -0,0 +1,101 @@
...
+          <div class='content-no-padding u-center'><div class={`divider ${classes.divider}`} /></div>
+
+          <textarea placeholder='Description' value={description} onChange={handleDescriptionChange} />
+          <input class='form-group-input input-small' placeholder='Flag' value={flag} onChange={handleFlagChange} />
+
+          <div style='height: 15px' />
+
+          <div class='col-12' style='display: flex; justify-content: space-between;'>
+            <button class='form-group-btn btn-small btn-info'>Update</button>
+            <button class='form-group-btn btn-small btn-danger' onClick={handleDelete} >Delete</button>
```
ethanwu10, line 93: This needs a `type="button"` attribute to avoid it becoming a submit button

client/src/components/admin/problem.js
```diff
@@ -0,0 +1,101 @@
...
+  const [category, setCategory] = useState(problem.category)
+  const handleCategoryChange = useCallback(e => setCategory(e.target.value), [])
+
+  const [author, setAuthor] = useState(problem.author)
+  const handleAuthorChange = useCallback(e => setAuthor(e.target.value), [])
+
+  const [name, setName] = useState(problem.name)
+  const handleNameChange = useCallback(e => setName(e.target.value), [])
+
+  const handleUpdate = e => {
```
ethanwu10, line 44: This should be in a `useCallback` hook to avoid re-registering a new event handler each render

client/src/components/admin/problem.js
```diff
@@ -0,0 +1,101 @@
...
+        flag,
+        description,
+        category,
+        author,
+        name
+      }
+    })
+  }
+
+  const handleDelete = e => {
```
ethanwu10, line 59: `useCallback`

Also, maybe we should have a confirmation modal before deleting?

client/src/components/admin/problem.js
```diff
@@ -0,0 +1,101 @@
...
+              <input class='form-group-input input-small' placeholder='Author' value={author} onChange={handleAuthorChange} />
+            </div>
+          </div>
+
+          <div class='content-no-padding u-center'><div class={`divider ${classes.divider}`} /></div>
+
+          <textarea placeholder='Description' value={description} onChange={handleDescriptionChange} />
+          <input class='form-group-input input-small' placeholder='Flag' value={flag} onChange={handleFlagChange} />
+
+          <div style='height: 15px' />
```
ethanwu10, line 89: This style should be applied through jss, and the height should be specified in `rem` (`1rem`)

client/src/components/admin/problem.js
```diff
@@ -0,0 +1,101 @@
...
+
+          <div class='col-12' style='display: flex; justify-content: space-between;'>
```
ethanwu10, line 91: This style should be applied through jss

client/src/routes/challs.js
```diff
@@ -0,0 +1,126 @@
...
+    if (!solveIDs.includes(id)) {
+      setSolveIDs([...solveIDs, id])
+    }
+  }, [solveIDs])
+
+  const handleInvertShowSolved = useCallback(() => {
+    setShowSolved(!showSolved)
+  }, [showSolved])
+
+  const handleInvertCategoryStateFn = prop => () => {
```
ethanwu10, line 25: This is creating a new handler function every render, consider refactoring the checkboxes into components that take the category as a prop and then pass the prop into an updater function (the function returned by this closure would work, just remove the `() =>` and wrap in `useCallback`)

client/src/routes/admin/challs.js
```diff
@@ -0,0 +1,44 @@
...
+
+import { getChallenges } from '../../api/admin/challs'
+
+const Challenges = ({ classes }) => {
+  const [problems, setProblems] = useState([])
+
+  useEffect(() => {
+    document.title = `Challenges${config.ctfTitle}`
+
+    getChallenges()
```
ethanwu10, line 15: This should be in a separate `useEffect` call

client/src/components/admin/problem.js
```diff
@@ -0,0 +1,108 @@
...
+        category,
+        author,
+        name
+      }
+    })
+  }, [problem, flag, description, category, author, name])
+
+  const handleDelete = useCallback(e => {
+    e.preventDefault()
+
```
ethanwu10, line 42: We should have a confirmation modal before deleting.

client/src/components/admin/problem.js
```diff
@@ -0,0 +1,167 @@
...
+  modalBody: {
+    paddingTop: '0em !important' // reduce space between header and body
+  },
+  controls: {
+    display: 'flex',
+    justifyContent: 'center',
+    '& :first-child': {
+      marginLeft: '0em'
+    },
+    '& :last-child': {
```
ginkoid, line 19: It would be a bit more clean if we replaced these with individual selectors for the cancel and delete challenge buttons, in case we ever add more buttons.

client/src/components/admin/problem.js
```diff
@@ -0,0 +1,167 @@
...
+  }, [onDelete])
+
+  return (
+    <Modal {...{ open, onClose }}>
+      <div class='modal-header'>
+        <div class='modal-title'>Delete Challenge?</div>
+      </div>
+      <div class={`modal-body ${classes.modalBody}`}>
+        This is an irreversible action that permanently deletes the challenge and revokes all solves.
+        <div class={`${classes.controls}`}>
```
ginkoid, line 40:
```suggestion
        <div class={classes.controls}>
```

client/src/components/admin/problem.js
```diff
@@ -0,0 +1,167 @@
...
+    e.preventDefault()
+    onClose()
+  }, [onClose])
+  const wrappedOnDelete = useCallback((e) => {
+    e.preventDefault()
+    onDelete()
+  }, [onDelete])
+
+  return (
+    <Modal {...{ open, onClose }}>
```
ginkoid, line 34:
```suggestion
    <Modal open={open} onClose={onClose}>
```

client/src/routes/challs.js
```diff
@@ -0,0 +1,143 @@
...
+                <label class='form-ext-label' for='check1'>Show Solved</label>
+              </div>
+            </div>
+          </div>
+        </div>
+        <div class={`frame ${classes.frame}`}>
+          <div class='frame__body'>
+            <div class='frame__title title'>Categories</div>
+            {
+              Object.keys(categories).map(category => {
```
ginkoid, line 101:
```suggestion
              Object.keys(categories).map((category, checked) => {
```

client/src/routes/challs.js
```diff
@@ -0,0 +1,143 @@
...
+                return (
+                  <div key={category} class='form-ext-control form-ext-checkbox'>
+                    <input id={category} class='form-ext-input' type='checkbox' checked={categories[category]} onChange={handleCategoryCheckedChange} />
```
ginkoid, line 104:
```suggestion
                    <input id={category} class='form-ext-input' type='checkbox' checked={checked} onChange={handleCategoryCheckedChange} />
```

client/src/components/admin/problem.js
ethanwu10: This is already set up to allow for more buttons, the selectors are for targeting whichever button is first and last to remove extra margin from the outside of the buttons; it isn't targeting the cancel and delete buttons specifically, but instead their positions.

PR #209, opened 2020-04-20
Registration rework

chen-robert:
Added team member information to the profile page. 

Note that you will need to `yarn migrate up` before testing this PR.


client/src/components/profile/memberscard.js
```diff
@@ -0,0 +1,127 @@
...
+      <div class='btn-container u-vertical-center'>
+        <input onClick={handleDelete} type='submit' class='btn-tiny btn-danger u-no-margin' value='Delete' />
+      </div>
+    </div>
+  )
+}
+
+const MembersCard = withStyles({
+
+}, withToast(({ classes, division: originalDivision, toast }) => {
```
ethanwu10, line 30: Prefer the hook over the HOC

client/src/components/profile/memberscard.js
```diff
@@ -0,0 +1,127 @@
...
+
+    const division = e.target.value
+
+    updateAccount({ division })
+      .then(({ error, data }) => {
+        if (error) {
+          toast({ body: error, type: 'error' })
+        } else {
+          setDivision(data.division)
+          toast({ body: 'Division successsfully updated' })
```
ethanwu10, line 54:
```suggestion
          toast({ body: 'Division successsfully updated', type: 'success' })
```

client/src/components/profile/memberscard.js
```diff
@@ -0,0 +1,127 @@
...
+    addMember({
+      name, email, grade
+    })
+      .then(({ error, data }) => {
+        setButtonDisabled(false)
+
+        if (error) {
+          toast({ body: error, type: 'error' })
+        } else {
+          toast({ body: 'Team member successfully added' })
```
ethanwu10, line 75:
```suggestion
          toast({ body: 'Team member successfully added', type: 'success' })
```

client/src/components/profile/memberscard.js
```diff
@@ -0,0 +1,127 @@
...
+
+  const [buttonDisabled, setButtonDisabled] = useState(false)
+
+  const [division, setDivision] = useState(originalDivision)
+  const handleDivisionChange = useCallback(e => {
+    e.preventDefault()
+
+    const division = e.target.value
+
+    updateAccount({ division })
```
ethanwu10, line 48: Do we want to send the update to the server every time theuser’s selection changes, or should it have an “update” button?

server/api/users/members/new.js
```diff
@@ -0,0 +1,48 @@
+const { v4: uuidv4 } = require('uuid')
+const database = require('../../../database')
+const { responses } = require('../../../responses')
+
+module.exports = {
+  method: 'post',
+  path: '/users/members/new',
```
ethanwu10, line 7: Semantically this should be a `POST` to `/users/members`

server/api/users/members/delete.js
```diff
@@ -0,0 +1,29 @@
...
+  },
+  handler: async ({ req, user }) => {
+    const { id } = req.params
+
+    await database.members.removeMember({
+      id,
+      userid: user.id
+    })
+
+    return responses.goodMemberCreate
```
ethanwu10, line 27:
```suggestion
    return responses.goodMemberDelete
```

client/src/components/profile/memberscard.js
```diff
@@ -0,0 +1,127 @@
...
+const MemberRow = ({ id, name, setMembers }) => {
+  const handleDelete = useCallback(() => {
+    removeMember({ id })
+      .then(() => {
+        setMembers(members => members.filter(a => a.id !== id))
+      })
+  }, [id, setMembers])
+
+  return (
+    <div class='u-vertical-center' style='width: 100%; display: flex; justify-content: space-between;' key={id}>
```
ethanwu10, line 19: style this with JSS? (and drop the `u-vertical-center` as well since we might as well set `align-items` ourselves, given that we’re explicitly specifying half the styles applied by that utility class already)

ethanwu10:
Should members be editable?

chen-robert:
Ideally yes @ethanwu10, did not implement it in this PR though.

client/src/components/profile/memberscard.js
chen-robert: Cirrus 'success' styling looks really bad on toasts, so I'm just using the default to denote "success".

client/src/components/profile/memberscard.js
chen-robert: See above

client/src/components/profile/memberscard.js
chen-robert: There's no rate limit so it should be fine to just auto-update on change.

client/src/components/profile/memberscard.js
```diff
@@ -0,0 +1,140 @@
...
+      <p class='u-no-margin'>{name}</p>
+      <div class='btn-container u-vertical-center'>
+        <input onClick={handleDelete} type='submit' class='btn-tiny btn-danger u-no-margin' value='Delete' />
+      </div>
+    </div>
+  )
+})
+
+const MembersCard = withStyles({
+
```
ginkoid, line 40: We can remove this `withStyles` HOC.

server/api/users/members/delete.js
```diff
@@ -0,0 +1,29 @@
+const database = require('../../../database')
+const { responses } = require('../../../responses')
+
+module.exports = {
+  method: 'delete',
+  path: '/users/members/:id',
```
ginkoid, line 6: To be more restful, we should probably name this `/users/me/members/:id`.

server/api/users/members/list.js
```diff
@@ -0,0 +1,15 @@
+const database = require('../../../database')
+const { responses } = require('../../../responses')
+
+module.exports = {
+  method: 'get',
+  path: '/users/members/',
```
ginkoid, line 6: To be more restful, we should probably name this `/users/me/members`.

server/api/users/members/new.js
```diff
@@ -0,0 +1,48 @@
+const { v4: uuidv4 } = require('uuid')
+const database = require('../../../database')
+const { responses } = require('../../../responses')
+
+module.exports = {
+  method: 'post',
+  path: '/users/members',
```
ginkoid, line 7: Same as above

client/src/components/profile/memberscard.js
chen-robert: I think it's fine to leave in, to make it easier to style in future.

PR #218, opened 2020-04-27
CTFtime integration improvements

chen-robert:
Resolved 3 small issues regarding CTFtime integration on frontend.


server/auth/register.js
```diff
@@ -22,7 +22,7 @@ const register = async ({ division, email, name, ctftimeId }) => {
       return responses.badKnownEmail
     }
     if (e.constraint === 'users_name_key') {
-      return responses.badKnownName
+      return [responses.badKnownName, name]
```
ginkoid, line 25: In future, we could return the team id and name from the `/integrations/ctftime/callback` endpoint so that other parts of the platform which use CTFtime auth can access the data in the CTFtime token.

client/src/routes/profile.js
```diff
@@ -184,51 +204,81 @@ const UpdateCard = withStyles({
...
+      }
+    }
+  }, [name, email, oldName, oldEmail, onUpdate, toast])
 
   return (
     <div class='card u-flex u-flex-column'>
       <div class='content'>
         <p>Update Information</p>
-        <p class='font-thin u-no-margin'>This will change how your team appears on the scoreboard. Note that you may only update your team's information once every 10 minutes.</p>
+        <p class='font-thin u-no-margin'>This will change how your team appears on the scoreboard. Note that you may only change your team's name once every 10 minutes (changing only email won't count).</p>
```
ginkoid, line 277:
```suggestion
        <p class='font-thin u-no-margin'>This will change how your team appears on the scoreboard. Note that you may only change your team's name once every 10 minutes.</p>
```

imo it's fine without the stuff in parenthesis.

PR #228, opened 2020-04-30
Added database provider

jimmyl02:
Wrote new challenge provider which interfaces with database
Created new challenges table

Maybe need to refactor database to ts in order to use imports


server/providers/challenges/database/index.ts
```diff
@@ -0,0 +1,86 @@
...
+class DatabaseProvider extends EventEmitter implements Provider {
+  private _updateInterval: number
+  private _interval: NodeJS.Timeout
+  private challenges: Challenge[]
+
+  constructor (options: DatabaseProviderOptions) {
+    super()
+    options = {
+      updateInterval: 60 * 1000,
+      // rDeployDirectory: ''
```
ethanwu10, line 20:
```suggestion
```

server/providers/challenges/database/index.ts
```diff
@@ -0,0 +1,86 @@
...
+      description: chall.description,
+      files: chall.files,
+      author: chall.author,
+      category: chall.category,
+      points: chall.points,
+      flag: chall.flag
+    })
+
+    if (updateResponse.length === 0) {
+      // Nothing was updated which means chall with id does not exist, create the challenge
```
ethanwu10, line 59: We should do this as a single upsert operation instead of retrying if the `UPDATE` failed

migrations/1587950739709_challenge-data.js
```diff
@@ -0,0 +1,16 @@
+exports.up = function (pgm) {
+  pgm.createTable('challenges', {
+    id: { type: 'uuid', primaryKey: true },
+    name: { type: 'string', unique: true, notNull: true },
+    description: { type: 'string' },
+    files: { type: 'text[]' },
+    author: { type: 'string' },
+    category: { type: 'string' },
+    points: { type: 'integer', notNull: true },
+    flag: { type: 'string', notNull: true }
```
ethanwu10, line 10: We may want to store the challenge data as a JSON blob in the database so that we can also support additional challenge metadata that rCTF treats as opaque that other tools can utilize

ethanwu10:
Tests are broken because the test setup uses the configs in the repository, which will result in no challenges from the database provider and thus a deadlock while the tests wait for challenges to load.

server/providers/challenges/database/index.ts
chen-robert: The `jsonb` format doesn't support upsert. However, there's no real implication here even if a race condition were possible. At worst, we use slightly outdated challenge data.

chen-robert:
I think `Promise<void>` is just for async functions, essentially the same since the return type isn't used.

PR #249, opened 2020-05-11
Added upload provider

chen-robert:
Upload provider and rework file storage.


chen-robert:
We should merge this before https://github.com/redpwn/rCTF/pull/228 because this PR standardizes the data stored.

server/providers/challenges/rdeploy-blob/index.ts
```diff
@@ -17,30 +31,89 @@ class RDeployBlobProvider extends EventEmitter implements Provider {
...
-      ? config.rDeployDirectory
-      : options.rDeployDirectory
+    this._rDeployDirectory = path.join(__dirname, '../../../../', options.rDeployDirectory)
     this._interval = setInterval(() => this._update(), this._updateInterval)
-    this._update()
+
+    this.downloadMap = new Map()
+
+    const fileDir = path.join(this._rDeployDirectory, options.rDeployFiles)
+    fs.readdir(fileDir)
```
ginkoid, line 51: Can we use async/await to reduce nesting here?

server/providers/challenges/rdeploy-blob/index.ts
```diff
@@ -17,30 +31,89 @@ class RDeployBlobProvider extends EventEmitter implements Provider {
...
+      .then(files => {
+        return Promise.all(
+          files.map(
+            file => {
```
ginkoid, line 55: We should filter to remove files which start with `.`, in case `.DS_Store` or similar gets in the dir somehow.

server/providers/uploads/local/index.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  private uploadMap: Map<string, Upload>
+
+  constructor (options: LocalProviderOptions, app: express.Application) {
+    if (options.uploadDirectory === undefined) {
+      options.uploadDirectory = path.join(process.cwd(), 'uploads')
+    }
+
+    if (!fs.existsSync(options.uploadDirectory)) {
+      fs.mkdirSync(options.uploadDirectory)
```
ginkoid, line 32: We should either use the `mkdirp` module or just not create the dir at all.

server/providers/challenges/rdeploy-blob/index.ts
```diff
@@ -17,30 +31,89 @@ class RDeployBlobProvider extends EventEmitter implements Provider {
...
+              const filePath = path.join(fileDir, file)
+
+              return fs.stat(filePath)
```
ginkoid, line 58: We got the list of files from a `readdir`, so this `stat` doesnt seem necessary.

server/providers/uploads/local/index.ts
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    this.uploadDirectory = options.uploadDirectory
+    this.endpoint = options.endpoint
+    this.app = app
+
+    this.uploadMap = new Map()
+    this.app.get(this.endpoint, (req, res) => {
```
ethanwu10, line 40: (style) Maybe define this as an instance method instead of as an anonymous function?

server/providers/uploads/local/index.ts
```diff
@@ -0,0 +1,72 @@
...
+      const key = req.query.key.toString()
+
+      if (this.uploadMap.has(key)) {
+        const upload = this.uploadMap.get(key)
+
+        res.download(upload.filePath, upload.name)
+      } else {
+        res.status(400)
```
ethanwu10, line 48: Why is this a 400 and not a 404?

server/providers/uploads/local/index.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    this.uploadDirectory = options.uploadDirectory
+    this.endpoint = options.endpoint
+    this.app = app
+
+    this.uploadMap = new Map()
+    this.app.get(this.endpoint, (req, res) => {
+      const key = req.query.key.toString()
+
+      if (this.uploadMap.has(key)) {
```
ethanwu10, line 43: Do we need to care if the file with a given hash was uploaded with the same name as it’s currently being requested as? I think a check for the file by hash existing on disk is good enough, and we can serve it under whatever name it was requested with.

server/providers/challenges/rdeploy-blob/index.ts
ethanwu10: Does rDeploy never create dotfiles? This gets complicated when there are dotfiles that we actually want to serve (e.g. challenge where a `.bash_history` is involved)

ethanwu10:
We may want to standardize the path used for uploads so that a tool can easily check if a certain file (by hash) already has been uploaded. Also, I think that having the hash and name be part of the path is preferable to having it in a query string parameter (e.g. `${endpoint}/:hash/:name`)

server/providers/challenges/rdeploy-blob/index.ts
chen-robert: If it's not referenced in a problem, there's no way to actually download the file. They'd have to guess the hash, which is essentially the same as guessing the contents of the file.

server/providers/challenges/rdeploy-blob/index.ts
chen-robert: We need to make sure it's not a directory or it might error

chen-robert:
That would probably be done in an API endpoitn

server/providers/uploads/local/index.ts
chen-robert: This would support the default installation of rctf where the uploads directory is just `/app/uploads`. I don't think we need to support weird custom configurations where the directory is nested somewhere deep.

server/uploads/index.ts
```diff
@@ -0,0 +1,19 @@
+import config from '../../config/server'
+import path from 'path'
+import express from 'express'
+import { Provider } from './types'
+
+let provider: Provider = null
+
+export const init = (app: express.Application): void => {
+  const ProviderClass = require(path.join('../providers', config.uploadProvider.name)).default
```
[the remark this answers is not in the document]
chen-robert: The initialization function should be synchronous to prevent race conditions on the registration of upload endpoint.

server/uploads/index.ts
```diff
@@ -0,0 +1,19 @@
...
+  provider = new ProviderClass(config.uploadProvider.options, app)
+}
+
+export const get = (): Provider => {
```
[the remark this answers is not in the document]
chen-robert: Seems like a lot of unnecessary boilerplate, but I agree we should standardize on one format. I'll make an issue for this.

server/providers/uploads/local/index.ts
chen-robert: That seems like unnecessary flexibility. Ideally, the key is an opaque string that maps directly to a file/name combo.

PR #253, opened 2020-05-15
feat(docker): add new multistage dockerfile build

ginkoid:
The new dockerfile works with K8S.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,31 +1,19 @@
-FROM node:12-slim
-
+FROM node:12.16.3-buster-slim AS build
 WORKDIR /app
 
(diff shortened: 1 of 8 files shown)


Dockerfile
```diff
@@ -1,31 +1,19 @@
...
-
-ENV PORT 8000
-EXPOSE 8000
-
-COPY . .
-
+RUN yarn
 
-CMD ["node", "dist/server/index.js"]
+CMD ["node", "/app/dist/server/index.js"]
```
ethanwu10, line 19: The `WORKDIR /app` should probably be kept in the final image

PR #255, opened 2020-05-18
Render mustache in WDS (yarn dev)

ethanwu10:
When in dev mode, render the Mustache templates in `index.html` within Webpack; this fixes `yarn dev`.

Currently the context for the mustache template is copy-pasted from the server rendering code.

### DIFF
--- a/preact.config.js
+++ b/preact.config.js
@@ -1,8 +1,12 @@
 require('dotenv').config()
 
+const path = require('path')
 const glob = require('glob')
+const assert = require('assert').strict
 
 const PurgecssPlugin = require('purgecss-webpack-plugin')
+
+const clientConfig = require('./config/client.js')
 
 export default (config, env, helpers) => {
   if (env.production) {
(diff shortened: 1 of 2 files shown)


preact.config.js
```diff
@@ -61,4 +67,21 @@ export default (config, env, helpers) => {
...
+      assert(/^!!ejs-loader!/.test(options.template))
+
+      // FIXME: refactor this (copy-pasted from server)
+      const context = {
+        config: JSON.stringify(clientConfig),
+        ctfName: clientConfig.ctfName,
+        meta: clientConfig.meta
+      }
+
+      options.template = `!!ejs-loader!mustache-rendered-loader?${JSON.stringify(context)}!${options.template.substr(13)}`
```
chen-robert, line 84: Where did `13` come from?

preact.config.js
```diff
@@ -61,4 +67,21 @@ export default (config, env, helpers) => {
       }
     }
   }
+
+  if (!env.production) {
```
chen-robert, line 71: Would prefer a feature flag for this. We don't want our builds to inject config if say, we accidently forget to set NODE_ENV.

ethanwu10: [preact-cli docs](https://github.com/preactjs/preact-cli/wiki/Config-Recipes#conditional-changes-based-on-env) seem to suggest using `env.production` to test whether or not the cli was invoked as `dev` or `build`.

Issue #252, opened 2020-05-15
Refactor Configuration


chen-robert:
The current configuration setup is very messy, being split between env variables and javascript files. 

In particular, we should dynamically load the start and end times. We should also symlink `/app/dist/config/` to `/app/config` to avoid mounting at `/app/dist` in production.

chen-robert:
Also, make difference between `serverOrigin` and `RCTF_ORIGIN` more explicit. One is for server while another is for client.

PR #266, opened 2020-05-18
Cleanup config

chen-robert:
Partially resolves https://github.com/redpwn/rCTF/issues/252. 

We still need to symlink in Docker build.


server/providers/uploads/local/index.ts
```diff
@@ -29,7 +29,7 @@ export default class LocalProvider implements Provider {
     }
 
     if (!fs.existsSync(options.uploadDirectory)) {
-      fs.mkdirSync(options.uploadDirectory)
+      fs.mkdirSync(options.uploadDirectory, { recursive: true })
     }
```
ethanwu10, line 33: The `exists` check is not necessary with `recursive: true`

PR #281, opened 2020-05-21
feat(docker): make docker compose deployment work

ginkoid:


.env.example
```diff
@@ -1,11 +1,8 @@
...
-RCTF_DATABASE_URL=postgresql://postgres:example_postgres_password@postgres:5432/postgres
-RCTF_REDIS_URL=redis://@redis:6379/0
-RCTF_SMTP_URL=smtp://user:password@smtp.server:587
-RCTF_EMAIL_FROM=example@example.com
+RCTF_TOKEN_KEY=32_byte_long_base64_encoded_value_for_token
+RCTF_ORIGIN=http://127.0.0.1:8000
+RCTF_DATABASE_PASSWORD=example_postgres_password
 RCTF_CTFTIME_CLIENT_ID=0
 RCTF_CTFTIME_CLIENT_SECRET=abcd
+RCTF_START_TIME=ms_since_epoch
```
chen-robert, line 7: Is it possible to provide sane values and have a comment? So it will work out of the box.

PR #295, opened 2020-05-28
Fix admin challs page

chen-robert:
Resolves https://github.com/redpwn/rCTF/issues/289


client/src/components/admin/problem.js
```diff
@@ -127,6 +159,31 @@ const Problem = ({ classes, problem }) => {
...
+            {
+              problem.files.length !== 0 &&
+                <div>
+                  <p class='faded frame__subtitle u-no-margin'>Downloads</p>
+                  <div class='tag-container'>
+                    {
+                      problem.files.map(file => {
+                        return (
+                          <div class='tag' key={file.url}>
+                            <a native download href={`${file.url}`}>
```
ginkoid, line 171:
```suggestion
                            <a native download href={file.url}>
```

client/src/routes/admin/challs.js
```diff
@@ -14,16 +14,19 @@ const SAMPLE_PROBLEM = {
...
-    max: 0
+    min: 100,
+    max: 500
   }
 }
 
 const Challenges = ({ classes }) => {
   const [problems, setProblems] = useState([])
-  const [newId, setNewId] = useState(uuid())
+  const newId = useMemo(() => uuid(), [problems])
```
ethanwu10, line 24:
```suggestion
  // eslint-disable-next-line react-hooks/exhaustive-deps
  const newId = useMemo(() => uuid(), [problems])
```
Potentially add an explanatory comment as well

server/api/admin/challs/put.js
```diff
@@ -21,17 +23,36 @@ export default {
       type: 'object',
       properties: {
         data: {
-          type: 'object'
+          type: 'object',
+          properties: {
+            files: {
+              type: 'array'
```
ginkoid, line 29: We need to validate that the elements in the array are strings.

client/src/components/admin/problem.js
```diff
@@ -68,20 +69,49 @@ const Problem = ({ classes, problem }) => {
...
   const handleNameChange = useCallback(e => setName(e.target.value), [])
 
-  const handleUpdate = useCallback(e => {
+  const [minPoints, setMinPoints] = useState(problem.points.min)
+  const handleMinPointsChange = useCallback(e => setMinPoints(e.target.value), [])
+
+  const [maxPoints, setMaxPoints] = useState(problem.points.max)
+  const handleMaxPointsChange = useCallback(e => setMaxPoints(e.target.value), [])
+
+  const fileElem = createRef()
```
ethanwu10, line 78: This should be `useRef()`

server/api/admin/challs/put.js
```diff
@@ -21,17 +23,36 @@ export default {
...
   handler: async ({ req }) => {
+    const uploadProvider = getUploadProvider()
     const chall = req.body.data
 
     // Ensure id is consistent
     chall.id = req.params.id
 
+    const files = await Promise.all(
+      chall.files.map(async ({ name, data }) => {
+        const url = await uploadProvider.upload(toBuffer(data))
```
ginkoid, line 45: We should catch any errors from `toBuffer` as it will throw if an invalid data URI is passed.

server/api/admin/challs/put.js
```diff
@@ -21,7 +21,24 @@ export default {
...
+                    type: 'string'
+                  },
+                  url: {
+                    type: 'string'
+                  }
+                },
+                required: ['name', 'url']
+              }
+            }
+          }
```
ginkoid, line 41: We should mark the `files` property as required.

server/api/admin/upload.js
```diff
@@ -0,0 +1,56 @@
...
+      required: ['files']
+    }
+  },
+  handler: async ({ req }) => {
+    const uploadProvider = getUploadProvider()
+
+    try {
+      const files = await Promise.all(
+        req.body.files.map(async ({ name, data }) => {
+          const url = await uploadProvider.upload(toBuffer(data))
```
ginkoid, line 42: Make sure to catch any errors from `toBuffer` because it throws from a malformed URI.

server/api/admin/challs/put.js
chen-robert: It's a partial update so there are no required properties

ethanwu10: This endpoint does a patch / merge, `files` must *not* be required to allow for a patch which does not change the files to succeed. We may wish to implement a separate schema validating challenges as they are written to the database to make sure they conform to the TypeScript interface definition.

server/api/admin/upload.js
ethanwu10: While this does catch errors in `toBuffer` due to the enclosing try/catch, we should treat errors thrown from `toBuffer` separately from errors thrown from the upload provider (which should 500).

PR #305, opened 2020-05-30
feat(server): split leaderboard into separate containers

ginkoid:


config/server.js
```diff
@@ -35,7 +35,7 @@ const config = {
       password: process.env.RCTF_REDIS_PASSWORD,
       database: process.env.RCTF_REDIS_DATABASE
     },
-    migrate: process.env.RCTF_DATABASE_MIGRATE || 'never'
+    migrate: process.env.RCTF_DATABASE_MIGRATE || 'never' // enum: never, before, only
```
chen-robert, line 38: We might want to move these into actual enum types at some point

PR #318, opened 2020-06-01
fix(server): fix raw index html being accessible

ethanwu10:


chen-robert:
We should make this work for `///` as well.

PR #320, opened 2020-06-01
feat(client): support ctftime modifcations for existing users

ginkoid:


client/src/components/profile/ctftime-card.js
```diff
@@ -0,0 +1,59 @@
...
+    const { kind, message } = await putCtftime({ ctftimeToken })
+    if (kind !== 'goodCtftimeAuthSet') {
+      toast({ body: message, type: 'error' })
+      return
+    }
+    onUpdate({ ctftimeId })
+  }, [toast, onUpdate])
+
+  const handleRemoveClick = useCallback(async () => {
+    const { kind, message } = await deleteCtftime()
```
ethanwu10, line 27: Would we want a confirmation dialog before deleting?

PR #363, opened 2020-06-16
feat(api): add tiebreakEligible flag for challenges

ginkoid:


test/unit/util.js
```diff
@@ -2,14 +2,8 @@ const test = require('ava')
...
-  const score = util.scores.getScore('static', 100, 100, 20)
-
-  t.is(score, 100)
-})
-
 test('get score dynamic', t => {
-  const score = util.scores.getScore('dynamic', 100, 500, 20)
-
-  t.is(score, 484)
+  t.is(util.scores.getScore(100, 500, 0), 500)
```
chen-robert, line 6: Do we want to hardcode test values? Or just make sure that the returned value seems reasonable, ie returned score is positive. Otherwise, our tests essentially dictate a hardcoded formula.

PR #389, opened 2020-07-02
docs: add in-tree docs

ethanwu10:
Set up configuration for MkDocs documentation in-tree


ethanwu10:
Currently RTD's version switcher does not work because it expects jQuery on the page.

PR #458, opened 2020-08-04
feat(config): use conf.d for unified config

ginkoid:


.github/workflows/ci.yml
```diff
@@ -69,19 +69,21 @@ jobs:
...
+        cp test/conf-test.yaml conf.d
 
     - name: Run migrations
       run: |
-        yarn migrate up
+        yarn migrate
+
+    - name: Add data to DB
+      run: |
+        psql postgres://postgres:password@localhost/rctf -c $'INSERT INTO challenges (id, data) VALUES (\'id\', \'{"flag": "flag{good_flag}", "name": "name", "files": [], "author": "author", "points": {"max": 500, "min": 100}, "category": "category", "description": "description", "tiebreakEligible": true}\')'
```
ethanwu10, line 86: Maybe use `$RCTF_DATABASE_URL` here?

install/config.sh
```diff
@@ -2,13 +2,19 @@
...
-cp -nR config/yml.example config/yml
+echo "ctfName: rCTF
+divisions:
+  open: Open
+origin: http://127.0.0.1:8000
+meta: 
+  description: 'A description of your CTF'
+  imageUrl: 'https://example.com'
+homeContent: 'A description of your CTF. Markdown supported.'
+tokenKey: '$(head -c 32 /dev/urandom | base64 -w0)'
```
ethanwu10, line 15: Since this is a secret, should this stay in .env?

package.json
```diff
@@ -19,7 +19,7 @@
     "test:slim": "ava",
     "test:report": "nyc --skip-full --reporter=lcov ava",
     "commit": "git cz",
-    "copy-static": "cpy \"server/**/*\" \"!**/*.{j,t}s\" \"migrations\" \"config/yml/*.yml\" dist/ --parents"
+    "copy-static": "cpy \"server/**/*\" \"!**/*.{j,t}s\" \"migrations\" \"conf.d\" dist/ --parents"
```
ethanwu10, line 22: IMO config shouldn’t be copied into dist/ tree, the config loading should read from repo root / directory above dist instead by default

server/config/client.ts
```diff
@@ -0,0 +1,40 @@
...
+  'ctfName' |
+  'divisions' |
+  'defaultDivision' |
+  'origin' |
+  'startTime' |
+  'endTime' |
+  'userMembers'
+> & {
+  emailEnabled: boolean;
+  ctftime: {
```
ethanwu10, line 17: Could be implemented as `Pick<ServerConfig['ctftime'], 'clientId'>`

Also missing nullable

server/config/server.ts
```diff
@@ -0,0 +1,206 @@
...
+
+const configPath = path.join(__dirname, '../../conf.d')
+const files = fs.readdirSync(configPath).filter(name => /\.(?:ya?ml|json)$/.test(name)).sort()
+const fileConfigs = files.map(name => yaml.parse(fs.readFileSync(path.join(configPath, name)).toString()))
+
+const parseBoolEnv = (val?: string): boolean | null => {
+  if (val == null) {
+    return null
+  }
+  return ['true', 'yes', 'y'].includes(val.toLowerCase().trim())
```
ethanwu10, line 90: Consider adding `1` to the list

server/config/server.ts
```diff
@@ -0,0 +1,206 @@
...
+    maxLimit: number;
+    maxOffset: number;
+    updateInterval: number;
+    graphMaxTeams: number;
+    graphSampleTime: number;
+  }
+  loginTimeout: number;
+}
+
+const configPath = path.join(__dirname, '../../conf.d')
```
ethanwu10, line 82: This should be settable via an env var

server/config/server.ts
```diff
@@ -0,0 +1,206 @@
...
+const files = fs.readdirSync(configPath).filter(name => /\.(?:ya?ml|json)$/.test(name)).sort()
+const fileConfigs = files.map(name => yaml.parse(fs.readFileSync(path.join(configPath, name)).toString()))
```
ethanwu10, line 84: Consider using different parsing methods based on file type (e.g. so yaml in a .json file will error)

server/config/server.ts
```diff
@@ -0,0 +1,206 @@
...
+
+const parseBoolEnv = (val?: string): boolean | null => {
+  if (val == null) {
+    return null
+  }
+  return ['true', 'yes', 'y'].includes(val.toLowerCase().trim())
+}
+
+const envConfig: ServerConfig = {
```
ethanwu10, line 93: This should be defined as a deep partial; otherwise this will fail strict null checks.

server/config/server.ts
```diff
@@ -0,0 +1,206 @@
...
+  homeContent: process.env.RCTF_HOME_CONTENT,
+  ctfName: process.env.RCTF_NAME,
+  meta: {
+    description: process.env.RCTF_META_DESCRIPTION,
+    imageUrl: process.env.RCTF_IMAGE_URL
+  },
+  logoUrl: process.env.RCTF_LOGO_URL,
+  globalSiteTag: process.env.RCTF_GLOBAL_SITE_TAG,
+  challengeProvider: {
+    name: undefined,
```
ethanwu10, line 128: What is the purpose of these `undefined` keys?

server/config/server.ts
```diff
@@ -0,0 +1,206 @@
...
+    maxLimit: 100,
+    maxOffset: 4294967296,
+    updateInterval: 10000,
+    graphMaxTeams: 10,
+    graphSampleTime: 1800000
+  },
+  loginTimeout: 3600000
+}
+
+const _removeUndefined = <T>(o: Record<string, T>): Record<string, T> | undefined => {
```
ethanwu10, line 183: These utility functions should be defined in a separate file so they can be unit-tested without executing config loading code.

server/providers/uploads/local/index.ts
```diff
@@ -29,6 +29,7 @@ export default class LocalProvider implements Provider {
   private uploadMap: Map<string, Upload>
 
   constructor (options: LocalProviderOptions, app: FastifyInstance) {
+    options = options ?? {}
```
ethanwu10, line 32: This should be handled when the backend is loaded. The argument to the constructor is not nullable, so from a TS perspective this is redundant. The provider loader should handle making sure the providers do not receive `undefined` as an argument.

server/responses.ts
```diff
@@ -1,3 +1,19 @@
...
+
+interface NormalResponseType extends BaseResponseType {
+  message: string
+}
+
+interface RawResponseType extends BaseResponseType {
+  rawContentType: string
+}
+
+type ResponseType = MergeExclusive<NormalResponseType, RawResponseType>
```
ethanwu10, line 15: Type is unused; this should be exported. Also consider type-checking `responseList` (perhaps by first declaring a temporary variable and then constructing a mapping type using `keyof`)

test/integration/profile.js
```diff
@@ -36,8 +38,9 @@ test.serial('succeeds with goodUserData', async t => {
   t.is(resp.body.data.name, testUser.name)
 })
 
-test.after.always('cleanup test user', async t => {
+test.serial.after('cleanup test user', async t => {
```
ethanwu10, line 41: Why was `always` removed?

server/config/server.ts
```diff
@@ -79,15 +79,29 @@ export type ServerConfig = {
...
+
+const configPath = path.resolve(__dirname, '../../..', process.env.RCTF_CONF_PATH ?? 'conf.d')
+const fileConfigs: PartialDeep<ServerConfig>[] = []
+fs.readdirSync(configPath).sort().forEach((name) => {
+  const matched = name.match(/\.(.+)$/)
+  if (matched === null || !fileConfigLoaders.has(matched[1])) {
+    return
+  }
+  const loader = fileConfigLoaders.get(matched[1])
+  const config = loader(fs.readFileSync(path.join(configPath, name)).toString())
```
ethanwu10, line 97: Consider casting to `PartialDeep<ServerConfig>`, or defining the loaders to return the config type

PR #469, opened 2020-08-08
refactor(server): move most server code to be self-contained

ethanwu10:
Move configuration for TypeScript, configuration for NYC, and tests to inside of the server/ directory.

Also remove build commands from `migrate` and `test` scripts in package.json


ethanwu10:
This PR is not worth the effort to rebase, since `next` will be promoted to `master` without merging.

PR #500, opened 2020-08-19
refactor(server): refactor all cache and leaderboard to use TS

ginkoid:


packages/server/src/cache/leaderboard.ts
```diff
@@ -0,0 +1,225 @@
...
+  }
+  const redisResult = await client.rctfGetRange(
+    getLeaderboardKey(division),
+    all ? 0 : (start as number) * 3,
+    all ? -1 : (end as number) * 3 - 1
+  )
+  const result = []
+  for (let i = 0; i < redisResult.length - 1; i += 3) {
+    // format the flat redis list response into an array of arrays
+    // i is the user id, i + 1 is the user name, i + 2 is the user score
```
ethanwu10, line 88: Outdated / unnecessary comment?

packages/server/src/cache/leaderboard.ts
```diff
@@ -0,0 +1,225 @@
...
+    maxTeams.toString()
+  )
+  if (redisResult === null) {
+    return []
+  }
+  const parsed = JSON.parse(redisResult) as string[]
+  const lastUpdate = parseInt(parsed[0])
+  const latest = parsed[1]
+  const graphData = parsed[2]
+  const result = []
```
ethanwu10, line 195: This should be typed.
```suggestion
  const result: ExternalGraphEntry[] = []
```

packages/server/src/cache/login.ts
```diff
@@ -0,0 +1,11 @@
+import config from '../config/server'
+import client from './client'
+
+export const makeLogin = async ({ id }: { id: string }): Promise<void> => {
```
ethanwu10, line 4: Should this be typed with `VerifyTokenData['verifyId']`?

packages/server/src/cache/login.ts
```diff
@@ -0,0 +1,11 @@
...
+  await client.set(`login:${id}`, '0', 'px', config.loginTimeout)
+}
+
+export const useLogin = async ({ id }: { id: string }): Promise<boolean> => {
```
ethanwu10, line 8: See above

packages/server/src/cache/timeouts.ts
```diff
@@ -0,0 +1,55 @@
...
+
+const typePrefixes = {
+  FLAG: 'FLAG'
+}
+
+export enum types {
+  UPDATE_PROFILE = 'UPDATE_PROFILE'
+}
+
+type PrefixType = Opaque<string>
```
ethanwu10, line 18: `Opaque` type should always have a const enum tag.

packages/server/src/cache/timeouts.ts
```diff
@@ -0,0 +1,55 @@
...
+  type,
+  userid,
+  duration,
+  limit
+}: {
+  type: RateLimitType,
+  userid: string,
+  duration: number,
+  limit: number
+}): Promise<{ ok: boolean, timeLeft: number | null }> => {
```
ethanwu10, line 44: This could be typed with a `MergeExclusive` for better DX. Also, the argument type and return type are likely better defined in interfaces rather than inline.

packages/server/src/leaderboard/calculate.ts
```diff
@@ -0,0 +1,143 @@
...
+    let maxSolves = 0
+    for (const [, challengeInfo] of challengeInfos) {
+      if (challengeInfo.solves > maxSolves) {
+        maxSolves = challengeInfo.solves
+      }
+    }
+
+    for (let i = 0; i < challenges.length; i++) {
+      const challenge = challenges[i]
+      const challengeInfo = challengeInfos.get(challenge.id) as ChallengeInfo
```
ethanwu10, line 78: This is more clearly expressed as a non-null assertion + an eslint disable comment. Also, why is this not implemented as a loop over `challengeInfos` to avoid nulls entirely?

packages/server/src/leaderboard/types.ts
```diff
@@ -0,0 +1,49 @@
...
+  config: ServerConfig
+}
+
+export type UserInfo = Pick<User, 'id' | 'name' | 'division'> & {
+  score: number,
+  lastSolve: number,
+  solvedChallengeIds: Challenge['id'][]
+}
+
+export type ExternalUserInfo = Pick<UserInfo, 'id' | 'name' | 'score'>
```
ethanwu10, line 20: I think the internal type should be the one with a prefix, since only the external type is used outside of this "module," and it seems awkward to be referencing an ExternalFoo as a consumer of the module.

packages/server/src/cache/timeouts.ts
```diff
@@ -0,0 +1,55 @@
...
+* Otherwise, the method will resolve to an object with the `ok` key set to true.
+*/
+export const checkRateLimit = async ({
+  type,
+  userid,
+  duration,
+  limit
+}: {
+  type: RateLimitType,
+  userid: string,
```
ethanwu10, line 41: This should be derived from the `User` type
```suggestion
  userid: User['id'],
```

packages/server/src/cache/client.ts
```diff
@@ -0,0 +1,43 @@
...
+    dropBufferSupport: true
+  }) as RedisClient
+} else {
+  client = new Redis({
+    host: creds.host,
+    port: creds.port,
+    password: creds.password,
+    db: creds.database,
+    dropBufferSupport: true
+  }) as RedisClient
```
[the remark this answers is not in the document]
ginkoid: Destructuring `creds` and then constructing the options object seems much more verbose than the current code, especially because it would require slightly special handling of `db` / `database`.

packages/server/src/leaderboard/calculate.ts
ginkoid: `points.min` and `points.max` are not available on `ChallengeInfo`, only on `Challenge`. Because it's not  possible to efficiently retrieve the `Challenge` for a given `ChallengeInfo`, we iterate over `challenges` instead of `challengeInfos`.

packages/server/src/cache/timeouts.ts
```diff
@@ -15,7 +16,8 @@ export enum types {
   UPDATE_PROFILE = 'UPDATE_PROFILE'
 }
 
-type PrefixType = Opaque<string>
+enum _prefixType {}
```
ethanwu10, line 19:
```suggestion
const enum _prefixType {}
```
(prevents code from being emitted)

PR #526, opened 2020-09-16
Added error logging for upload exceptions

captainGeech42:


ethanwu10 (approved):
lgtm, though we may want to improve the interface between upload providers and the rest of rCTF to allow for more granular errors (for instance to distinguish between a 4xx and a 5xx response).

PR #544, opened 2020-11-06
Put API route definitions and types into a shared package

ethanwu10:
Create the `@rctf/api-types` package, meant to be consumed by both clients and `@rctf/server`. This package is built by turning YAML route definitions into TypeScript type definitions & JS files via a code generator.

This PR also refactors all\* server route handlers to use these generated types and thus be type-safe.

\* Routes in the registration flow have not been converted to TS yet due to issues with typing their implementation - they are migrated to the new route declaration mechanism, but will remain as JS and will get typed in another PR.


ethanwu10:
The PR should be complete now, just needs a squash and review to be mergable.

packages/server/src/api/leaderboard/graph.js
```diff
@@ -1,65 +0,0 @@
...
-      type: 'object',
-      properties: {
-        division: {
-          type: 'string',
-          enum: Object.keys(config.divisions),
-        },
-        limit: {
-          type: 'integer',
-          minimum: 1,
-          maximum: config.leaderboard.graphMaxTeams,
```
ethanwu10, line 20: Currently, runtime schemas are missing important properties derived from config such as this. `makeFastifyRoute` should allow deepmerging on the schema using the third argument to enable this AJV-based parameter validation.
Routes must also be checked manually to make sure none of these types of constraints made it in to the static `@rctf/api-types` sources.

ethanwu10: I believe I have covered everything...

chen-robert (approved):

ginkoid (approved):

ethanwu10:
Some lingering potential improvements:
- Consider simplifying `params` schema format - implicit `type: object` + marking all properties required, also validation against detected params in the `path`
- Consider having shared definitions - needs extra tooling if they are to stay as references after codegen, or codegen can do the substitution

Merging

PR #547, opened 2020-12-02
fix(graph): limit graph label count for long ctfs

ginkoid:


DarinMao (approved):
how much we gotta pay you to use a graph library

PR #555, opened 2021-02-08
fix: small errors

0xflotus:


ethanwu10:
I'm dropping 71435a7 since all our other docs use `yaml`, so changing one fence to `yml` introduces inconsistency; also, the syntax highlighting works correctly on both the docs site mkdocs build and GitHub, so the change is redundant to begin with.
Rewriting c5dcdc7 to remove spurious issue reference.

ethanwu10 (approved):

PR #559, opened 2021-03-28
batch dependency upgrade + related fixes

ethanwu10:
Update to Webpack 5, Storybook 6.2 beta with webpack 5 builder, AJV 7.x (8.x was released 18 hours ago and will be handled another time because this PR took long enough as it is).

Important changes:
- react is added as devDep to client since Storybook has (mostly) committed to keeping react as a pure peerdep, and thus this is necessary for SB to work
- `size-plugin` held at 2.x because 3.x breaks on webpack 5
- storyshots snapshot tests removed for being too flaky and often not capturing significant details - 50% of snapshots changed with a theme-ui update, none of the emotion serializers seem to want to work with preact, the shallow rendering often did render the actual relevant parts of stories, and hooks in story functions seem to be broken

Other notes:
- consider "linting" api-types schemas using AJV in strict mode for diagnostics


chen-robert (approved):
lgtm

DarinMao (approved):
wandoze

PR #573, opened 2021-05-27
chore: backport ranking users with only tiebreakEligible: false

ethanwu10:
Backports changes from #543

Also adjusted types at the module boundary of the leaderboard worker code to hide implementation detail data.
This change also strips the data from the objects, resulting in the creation of a new JS object per leaderboard entry, but keeps snapshots cleaner. We may wish to consider doing this in only snapshot serialization in the future.

### DIFF
--- a/packages/server/src/leaderboard/calculate.ts
+++ b/packages/server/src/leaderboard/calculate.ts
@@ -35,7 +35,8 @@ export default ({
       name: user.name,
       division: user.division,
       score: 0,
-      lastSolve: 0,
+      lastSolve: undefined,
+      lastTiebreakEligibleSolve: undefined,
       solvedChallengeIds: [],
     })
   }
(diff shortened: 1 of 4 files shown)


chen-robert (approved):

packages/server/src/leaderboard/calculate.ts
```diff
@@ -122,14 +131,24 @@ export default ({
...
+    if (
+      a.lastTiebreakEligibleSolve !== undefined ||
+      b.lastTiebreakEligibleSolve !== undefined
+    ) {
+      return (
+        (a.lastTiebreakEligibleSolve ?? Infinity) -
+        (b.lastTiebreakEligibleSolve ?? Infinity)
+      )
+    }
+    return (a.lastSolve ?? Infinity) - (b.lastSolve ?? Infinity)
```
ethanwu10, line 151: This can result in the compare function returning `NaN`, which would make `sort` implementation-defined behavior according to spec. If both have undefined last solve times, we need to make sure to return 0.

ginkoid (approved):

